Issue #25, opened 2020-09-27
Add CI/CD


Derek-Wds:
Hi all, I suggest we should add CI/CD as soon as possible. It will be necessary to do so before we could bring more features into the project. Here are some most popular CI tools that I know:

* [Travis CI](https://docs.travis-ci.com/user/tutorial/)
* [Circle CI](https://circleci.com/docs/2.0/)

Github also provides a CI/CD tool: [GitHub Actions](https://github.com/microsoft/qlib/actions/new). It is another good alternative for us to set up a workflow.

Besides, we could add some issue templates that can help users make bug reports and feature requests more clearly.

Derek

PR #26, opened 2020-09-27
Add CI & Issue Template

Derek-Wds:
Hi, this PR is about adding CI to the library. [#25 ](https://github.com/microsoft/qlib/issues/25) I chose to use Github Action for the CI of our library.

This is an initial version of the CI for test. I will update it in the future if any error occurs.

Please help me double check it. Thanks!

Derek


Derek-Wds:
Hi, I have just double checked that `flake8` actually includes `PEP8` style check for python. Thus, I removed the `PEP8` command we added in the previous commit.

.github/ISSUE_TEMPLATE/bug-report.md
```diff
@@ -0,0 +1,38 @@
...
+## Screenshot
+
+<!-- A screenshot of the error message or anything shouldn't appear-->
+
+## Environment
+
+ - Qlib version:
+ - Python version:
+ - OS (`Windows`, `Linux`, `MacOS`):
+ - Commit number:
```
you-n-g, line 34: It would be better if we could provide script to collect these info. User could get these info with a single command.
`curl -sL https://xxxx/collect_info.py | python`
Example of this script
```python
import qlib
print(qlib.VERSION)
.......
```

you-n-g: This will help both users and us to collect information

you-n-g, line 34: Commit number is optional.
It is only required when user uses the dev version

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Description
+<!--- Describe your changes in detail -->
+
+## Motivation and Context
+<!--- Are there any related issues? If so, please put the link here. -->
+<!--- Why is this change required? What problem does it solve? -->
+
+## How Has This Been Tested?
+<!--- Please describe in detail how you tested your changes. -->
```
you-n-g, line 11: Could we make this section simpler?

1. Tell user the command to run the tests `test_all_pipeline.py` and  just ask if user has run it? 
2. How to run users new test (if appropriate)

.github/ISSUE_TEMPLATE/bug-report.md
Derek-Wds: Yeah, this might be a good solution to get qlib version. Maybe pengrong could help to upload such a file to the server.

Derek-Wds: Ok, I will update the information here.

.github/PULL_REQUEST_TEMPLATE.md
Derek-Wds: Ok, let me see how to tell users run the test script.

Issue #47, opened 2020-10-03
 How to use estimator_config_dnn.yaml without GPU ?


pan-cai:
If there is not GPU, 

```
MODEL_CONFIG = {
                "loss": "mse",
                'input_dim': 158,
                'output_dim': 1,
                'lr': 0.002,
                'lr_decay': 0.96,
                'lr_decay_steps': 100,
                'optimizer': 'adam',
                'max_steps': 8000,
                'batch_size': 4096,
                # 'GPU': '0',
                }
```

```
     73 def _check_driver():
     74     if not hasattr(torch._C, '_cuda_isDriverSufficient'):
---> 75         raise AssertionError("Torch not compiled with CUDA enabled")
     76     if not torch._C._cuda_isDriverSufficient():
     77         if torch._C._cuda_getDriverVersion() == 0:

AssertionError: Torch not compiled with CUDA enabled
```
----
```
MODEL_CONFIG = {
                "loss": "mse",
                'input_dim': 158,
                'output_dim': 1,
                'lr': 0.002,
                'lr_decay': 0.96,
                'lr_decay_steps': 100,
                'optimizer': 'adam',
                'max_steps': 8000,
                'batch_size': 4096,
                'GPU': '0',
                }
```
```
     73 def _check_driver():
     74     if not hasattr(torch._C, '_cuda_isDriverSufficient'):
---> 75         raise AssertionError("Torch not compiled with CUDA enabled")
     76     if not torch._C._cuda_isDriverSufficient():
     77         if torch._C._cuda_getDriverVersion() == 0:

AssertionError: Torch not compiled with CUDA enabled
```

bxdd:
Hi, thanks for your feedback. So far you can't run the baseline model without GPU. If your computer has multiple GPUs, the parameter "GPU" indicates which GPU to use instead of whether to use GPU. Please make sure that Cuda is installed on your computer before running the baseline model.

In the near future, we may make this example model compatible with CPU. Also, you can integrate your custom model into `Qlib`, and your PR will be welcome.

bxdd:
If you want to integrate custom models into `Qlib`, please refer to the Integration document(https://qlib.readthedocs.io/en/latest/start/integration.html) for more details.

bxdd:
Hi, We just now update the DNN model on PR #49. If Cuda is not installed on your computer, then `Qlib` will automatically use Cpu to train the model.

Looking forward to your further feedback.

pan-cai:
Wow, thank you so much

PR #49, opened 2020-10-04
Make the DNN model compatible with CPU & Update the Estimator doc

bxdd:
## Description
This pr aims to make the DNN model compatible with CPU and update the `Estimator` doc.

## Motivation and Context
The pr is related to #47 

## How Has This Been Tested?

## Types of changes

- [ ] Fix bugs
- [x] Add new feature
- [x] Update documentation

### DIFF
--- a/docs/component/estimator.rst
+++ b/docs/component/estimator.rst
@@ -574,10 +574,24 @@ The `qlib_data` field describes the parameters of qlib initialization.
       region: "cn"
     
 - `provider_uri`
-    The local directory where the data loaded by 'get_data.py' is stored.
+    Type: str. The URI of the Qlib data. For example, it could be the location where the data loaded by ``get_data.py`` are stored.
 - `region`
-    - If region == ``qlib.config.REG_CN``, 'qlib' will be initialized in US-stock mode. 
-    - If region == ``qlib.config.REG_US``, 'qlib' will be initialized in china-stock mode.
+    - If `region` == "us", ``Qlib`` will be initialized in US-stock mode. 
+    - If `region` == "cn", ``Qlib`` will be initialized in china-stock mode.
+- `redis_host`
+    Type: str, optional parameter(default: "127.0.0.1"), host of `redis`
+        The lock and cache mechanism relies on redis.
+- `redis_port`
+    Type: int, optional parameter(default: 6379), port of `redis`
+
+    .. note:: 
+        
+        The value of `region` should be aligned with the data stored in `provider_uri`. Currently, ``scripts/get_data.py`` only provides China stock market data. If users want to use the US stock market data, they should prepare their own US-stock data in `provider_uri` and switch to US-stock mode.
+
+    .. note::
+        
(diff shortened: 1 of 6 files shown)


docs/component/estimator.rst
```diff
@@ -574,10 +574,24 @@ The `qlib_data` field describes the parameters of qlib initialization.
       region: "cn"
     
 - `provider_uri`
-    The local directory where the data loaded by 'get_data.py' is stored.
+    Type: str. The URI of the Qlib data. For example, it could be the location where the data loaded by ``get_data.py`` are stored.
 - `region`
-    - If region == ``qlib.config.REG_CN``, 'qlib' will be initialized in US-stock mode. 
-    - If region == ``qlib.config.REG_US``, 'qlib' will be initialized in china-stock mode.
+    - If `region` == "us", ``Qlib`` will be initialized in US-stock mode. 
+    - If `region` == "cn", ``Qlib`` will be initialized in china-stock mode.
```
you-n-g, line 580: The concerete value in is upper case [here](https://github.com/microsoft/qlib/blob/main/qlib/config.py#L64)
I think change all of them to lower case will be better

qlib/contrib/model/pytorch_nn.py
```diff
@@ -47,7 +47,7 @@ def __init__(
         self,
         input_dim,
         output_dim,
-        layers=(256, 256, 128),
+        layers=(256, 512, 768, 1024, 768, 512, 256, 128, 64),
```
you-n-g, line 50: Will the change of NN structure affects the performance greatly?

qlib/contrib/model/pytorch_nn.py
```diff
@@ -133,11 +136,14 @@ def __init__(
         )
 
         self._fitted = False
-        self.dnn_model.cuda()
+
```
you-n-g, line 139: PEP8: too many blanks

you-n-g: @Derek-Wds   It seems that we have run the pep8 checker but the errors are not raised.
image

We use [black ](https://github.com/psf/black)with max line width set to 120 as our pep8 checker.

qlib/contrib/model/pytorch_nn.py
```diff
@@ -316,20 +330,20 @@ def update(self, val, n=1):
 
 
 class Net(nn.Module):
-    def __init__(self, input_dim, output_dim, layers=(256, 256, 256), loss="mse"):
+    def __init__(self, input_dim, output_dim, layers=(256, 512, 768, 512, 256, 128, 64), loss="mse"):
         super(Net, self).__init__()
         layers = [input_dim] + list(layers)
         dnn_layers = []
-        drop_input = nn.Dropout(0.1)
+        drop_input = nn.Dropout(0.05)
```
you-n-g, line 337: Will these changes of structure affect the NN performance greatly?

qlib/contrib/model/pytorch_nn.py
bxdd: Yes

PR #67, opened 2020-11-11
Fix re about provider_uri & Fix 'exit not defined' error in ipython

bxdd:
## Description
fix re about provider_uri & fix 'exit not defined' error in ipython

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/config.py
+++ b/qlib/config.py
@@ -211,14 +211,11 @@ class QlibConfig(Config):
             self["provider_uri"] = str(Path(self["provider_uri"]).expanduser().resolve())
 
     def get_uri_type(self):
-        win_rm = re.match("^[a-zA-Z]:.*", self["provider_uri"])
-        nfs_or_win_rm = re.match("^[^/]+:.+", self["provider_uri"])
-        # Windows path is "C:\\"
-        if win_rm is None:
-            if nfs_or_win_rm is None:
-                return QlibConfig.LOCAL_URI
-            else:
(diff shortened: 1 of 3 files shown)


qlib/config.py
```diff
@@ -211,12 +211,16 @@ def resolve_path(self):
             self["provider_uri"] = str(Path(self["provider_uri"]).expanduser().resolve())
 
     def get_uri_type(self):
-        rm = re.match("^[^/ ]+:.+", self["provider_uri"])
+        win_rm = re.match("^[a-zA-Z]:.*", self["provider_uri"])
+        nfs_or_win_rm = re.match("^[^/]+:.+", self["provider_uri"])
         # Windows path is "C:\\"
-        if rm is None or Path(self["provider_uri"]).exists():
-            return QlibConfig.LOCAL_URI
+        if win_rm is None:
```
you-n-g, line 217: The logic could be simpler
```python
win = re.match("^[a-zA-Z]:.*", self["provider_uri"]) is not None  # such as 'C:\\data', 'D:'
nfs_or_win = re.match("^[^/]+:.+", self["provider_uri"]) is not None  # such as 'host:/data/'   (User may define short hostname by themselves)

if nfs_or_win and not win:
     return QlibConfig.NFS_URI
else:
     return QlibConfig.LOCAL_URI
```

PR #113, opened 2020-12-13
Update the Sign operation in ops.py

YifanDengWHU:
Fix the bug when Sign followed by True/False value.

## Description

During the alpha generation exps, I found that when the Sign operation meets True/False value, it will raise a MaybeEncodingError. Here I transform the True/False value to np.float32.
## How Has This Been Tested?
- [x ] If you are adding a new feature, test on your own test scripts.

1. Pipeline test:
2. Your own tests:

## Types of changes

- [x ] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/data/ops.py
+++ b/qlib/data/ops.py
@@ -143,6 +143,11 @@ class Sign(ElemOperator):
     def __init__(self, feature):
         super(Sign, self).__init__(feature, "sign")
 
+    def _load_internal(self, instrument, start_index, end_index, freq):
+        series = self.feature.load(instrument, start_index, end_index, freq)
+        series = series.astype(np.float32)
+        return getattr(np, self.func)(series)
+
 
 class Log(ElemOperator):
     """Feature Log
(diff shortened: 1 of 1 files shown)


you-n-g:
Hi, 
Many thanks for sending the pull request.
But I can't understand why the data type conversion is required. Can you give us more details? Thanks

image

YifanDengWHU:
image
I mean when the sign function meets bool value, it will raise error. Here is the expression I test yesterday:
Sign(Gt($close, Sum(WMA($volume, 14),17)))
The $close value will be smaller than Sum(WMA($volume, 14),17)). So it is Sign(series([False, False, .......])) and it will raise error.
Though this expression is meaningless, but I think we should make it work?

you-n-g:
@YifanDengWHU  Thanks. I see now. 

Please add some comments to describe this case. 
In addition, please add this comments  above the `np.float32`
```python
# TODO:  The the precision should be configurable
```

Thanks for your contribution!

PR #183, opened 2021-01-07
Fix unrecognized config bug

fzc621:
The config key should be checked first instead of assignment, otherwise the if-condition would never be true.
## Description

Put the assignment statement after the if-statement.
## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation


you-n-g:
@fzc621  It's great! 
Thanks so much!

PR #197, opened 2021-01-18
fix setup error

GoooIce:
why required pymongo

```shell
pip install -U xxxxx

...
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.

pyqlib 0.6.1.dev0 requires pymongo==3.7.2, but you'll have pymongo 3.11.2 which is incompatible.
```

## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/scripts/collect_info.py
+++ b/scripts/collect_info.py
@@ -33,7 +33,6 @@ class InfoCollector:
             "scipy",
             "requests",
             "sacred",
-            "pymongo",
             "python-socketio",
             "redis",
             "python-redis-lock",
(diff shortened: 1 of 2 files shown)


you-n-g:
It's great!
Thanks so much!

PR #234, opened 2021-01-31
move `freq` params to dataloader

you-n-g:
move `freq` params to dataloader

## Description
The frequency should be tranparent to data handler.

1. Pipeline test:
2. Your own tests:

## Types of changes

- [ ] Fix bugs
- [X] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/data/data.py
+++ b/qlib/data/data.py
@@ -961,8 +961,7 @@ class BaseProvider:
         is a provider class.
         """
         disk_cache = C.default_disk_cache if disk_cache is None else disk_cache
-        if C.disable_disk_cache:
-            disk_cache = False
+        fields = list(fields)  # In case of tuple.
         try:
             return DatasetD.dataset(instruments, fields, start_time, end_time, freq, disk_cache)
         except TypeError:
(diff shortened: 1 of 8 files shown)


qlib/data/data.py
```diff
@@ -961,8 +961,7 @@ def features(
         is a provider class.
         """
         disk_cache = C.default_disk_cache if disk_cache is None else disk_cache
-        if C.disable_disk_cache:
```
you-n-g, line 964: @zhupr Please help to check the correctness of the modification

qlib/data/dataset/handler.py
```diff
@@ -71,38 +71,58 @@ def __init__(
...
 
         # Setup data loader
         assert data_loader is not None  # to make start_time end_time could have None default value
 
+        # what data source to load data
         self.data_loader = init_instance_by_config(
             data_loader,
             None if (isinstance(data_loader, dict) and "module_path" in data_loader) else data_loader_module,
             accept_types=DataLoader,
+            **kwargs,
```
bxdd, line 94: Is it necessary to pass kwargs to `data_loader`?
I think all arguments of `data_loader` should be in config.

bxdd: Is handlers in contrib should keep consistent with example? Pass arguments to data_loader with kwargs or config?

you-n-g:
@bxdd  I've removed `kwargs` now.
Please let me know if there are any other problems  :) w

Issue #225, opened 2021-01-27
How to load dnn models?


PayneJoe:
It seems we can not load model with R.load_object method? Do we have to implement the loading logic with our own?

By the way, I just tried the TFT model.

Thanks~

Derek-Wds:
Hi @PayneJoe , thanks for pointing this out! Yes, indeed, the `R` API we have in Qlib doesn't support saving models, and users have to implement the logic by themselves. 

Qlib has a base model API where users can create their own models. As you can see, the models can either be GBDT or DNN models, which may have different implementations based on various packages (Pytorch, Tensorflow, Lightgbm, etc.). This makes it difficult to find a compact and universal way to save all these models for now. 

We will try to figure something out in the future and if you have any good ideas, feel free to talk with us. Thanks!

PayneJoe:
> Hi @PayneJoe , thanks for pointing this out! Yes, indeed, the `R` API we have in Qlib doesn't support saving models, and users have to implement the logic by themselves.
> 
> Qlib has a base model API where users can create their own models. As you can see, the models can either be GBDT or DNN models, which may have different implementations based on various packages (Pytorch, Tensorflow, Lightgbm, etc.). This makes it difficult to find a compact and universal way to save all these models for now.
> 
> We will try to figure something out in the future and if you have any good ideas, feel free to talk with us. Thanks!

Hi @Derek-Chu , thanks for your immediate reply! 

It troubled me for few hours, since the connecting 'rope' between offline backtesting and actual conduction, either intra-day manual trading or automatic trading, is missing. But it's okay for non-dnn models. 

I am testing the inference part with loaded model TFT, and seeking a more beautiful solution for it is on my roadmap!

Thanks~

Derek-Wds:
We will support model saving and loading in the near future. I will reopen this issue for tracking the process.

Thanks~

Issue #211, opened 2021-01-23
model load from recorder.load_object() missing _fitted variable


2young-2simple-sometimes-naive:
## 🐛 Bug Description

model loaded from recorder.load_object() missing _fitted variable

## To Reproduce

1. `model = init_instance_by_config(task["model"])`, the model can be LSTM from pytorch
2. 
```
with R.start(experiment_name="train_model"):
    R.log_params(**flatten_dict(task))
    model.fit(dataset, save_path = ".model-pytorch")
    R.save_objects(trained_model=model)
    recorder = R.get_recorder()
```
3. 
```
with R.start(experiment_name="backtest_analysis"):
    recorder = R.get_recorder(rid, experiment_name="train_model")
    model = recorder.load_object("trained_model")

    # prediction
    recorder = R.get_recorder()
    ba_rid = recorder.id
    sr = SignalRecord(model, dataset, recorder)
    sr.generate()
```
It will return error saying _fitted not exist.
## Expected Behavior

recorder.load_object should be able to load all variables in LSTM class

## Environment

**Note**: User could run `cd scripts && python collect_info.py all` under project directory to get system information
and paste them here directly.

 - Qlib version:0.6.1.dev0
 - Python version: 3.7
 - OS (`Windows`, `Linux`, `MacOS`): Linux
 - Pytorch: 1.7.1+cu101

you-n-g:
Many objects are subclasses of `Serializable` https://github.com/microsoft/qlib/blob/main/qlib/model/base.py#L8
So only part of data will be dumped into pickle file.
If you want to dump more data for a specific instance,  you could override  some methods in this class

Thanks

PR #261, opened 2021-02-05
Update models to enable save/load

Derek-Wds:
## Description

This is related to the issues: [#211](https://github.com/microsoft/qlib/issues/211) and [#225](https://github.com/microsoft/qlib/issues/225). Users can directly save models through `R.save_objects`.

## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [x] Fix bugs
- [x] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/data/data.py
+++ b/qlib/data/data.py
@@ -478,13 +478,13 @@ class DatasetProvider(abc.ABC):
 
         data = pd.DataFrame(obj)
         _calendar = Cal.calendar(freq=freq)
-        data.index = _calendar[data.index.values.astype(np.int)]
+        data.index = _calendar[data.index.values.astype(int)]
         data.index.names = ["datetime"]
 
         if spans is None:
             return data
         else:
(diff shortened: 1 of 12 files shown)


qlib/data/data.py
```diff
@@ -478,13 +478,13 @@ def expression_calculator(inst, start_time, end_time, freq, column_names, spans=
 
         data = pd.DataFrame(obj)
         _calendar = Cal.calendar(freq=freq)
-        data.index = _calendar[data.index.values.astype(np.int)]
+        data.index = _calendar[data.index.values.astype(int)]
```
you-n-g, line 481: Why is this necessary ?

Derek-Wds: `np.int`, and `np.bool` is deprecated in numpy 1.20.0. We can use `int` directly or specify `np.int32` or `np.int64`.

PR #273, opened 2021-02-16
safe yaml loader

you-n-g:
## Description
For fixing https://github.com/microsoft/qlib/commit/802dac81c960b609d4e3212f55b26bd667e2a187

## Types of changes

- [X] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/__init__.py
+++ b/qlib/__init__.py
@@ -147,7 +147,7 @@ def init_from_yaml_conf(conf_path, **kwargs):
     """
 
     with open(conf_path) as f:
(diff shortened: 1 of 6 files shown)


wphan:
https://github.com/microsoft/qlib/search?q=yaml.load it looks like there are a couple more locations where `yaml.load` is used

wphan:
you might also want to use `yaml.safe_load` rather than setting the loader: https://pyyaml.org/wiki/PyYAMLDocumentation to be more explicit

you-n-g:
@vphan It is fixed now. Please check it :)

PR #310, opened 2021-03-05
high frequency example

javaThonc:
Submit a high-frequency example

## Description
Train under minute level data and test with signal test


examples/highfreq/high_freq_tree.ipynb
```diff
@@ -0,0 +1,192 @@
+{
```
you-n-g, line 1: Could you paste some experiments results in this folder?

you-n-g:
qrun example

PR #314, opened 2021-03-08
(1) Fix /0 bug in double_ensemble, (2) remove _default_uri for R/expm, (3) support model size in pytorch models

D-X-Y:
Four changes:
- Fix /0 bug in double_ensemble
- simplify the logic of default_uri for R/expm
- support model size in pytorch models

## Description
- In double_ensemble models, it might potentially divide zero and raise an error.
- Currently, _default_uri for R maybe inconsistent with that in C.
- In pytorch-based contrib codes, there are some useless kwargs, such as verbose.
- In pytorch-based contrib codes, add a function to count the number of parameters.

## Motivation and Context
Make it more friendly for researchers.

## How Has This Been Tested?

I run `pytest -W ignore::DeprecationWarning qlib/tests/test_all_pipeline.py`, below is the screenshot:
image

## Types of changes

- [x] Fix bugs
- [x] Add new feature

### DIFF
--- a/qlib/contrib/backtest/account.py
+++ b/qlib/contrib/backtest/account.py
@@ -104,10 +104,9 @@ class Account:
             # if suspend, no new price to be updated, profit is 0
             if trader.check_stock_suspended(code, today):
                 continue
-            else:
-                today_close = trader.get_close(code, today)
-                profit += (today_close - self.current.position[code]["price"]) * self.current.position[code]["amount"]
-                self.current.update_stock_price(stock_id=code, price=today_close)
+            today_close = trader.get_close(code, today)
+            profit += (today_close - self.current.position[code]["price"]) * self.current.position[code]["amount"]
+            self.current.update_stock_price(stock_id=code, price=today_close)
         self.rtn += profit
         # update holding day count
         self.current.add_count_all()
--- a/qlib/contrib/model/pytorch_utils.py
+++ b/qlib/contrib/model/pytorch_utils.py
@@ -4,7 +4,19 @@
 import torch.nn as nn
 
 
-def count_parameters(models_or_parameters, unit="mb"):
+def count_parameters(models_or_parameters, unit="m"):
+    """
+    This function is to obtain the storage size unit of a (or multiple) models.
+
+    Parameters
+    ----------
+    models_or_parameters : PyTorch model(s) or a list of parameters.
+    unit : the storage size unit.
+
+    Returns
+    -------
+    The number of parameters of the given model(s) or parameters.
+    """
     if isinstance(models_or_parameters, nn.Module):
         counts = sum(v.numel() for v in models_or_parameters.parameters())
     elif isinstance(models_or_parameters, nn.Parameter):
@@ -13,12 +25,13 @@ def count_parameters(models_or_parameters, unit="mb"):
         return sum(count_parameters(x, unit) for x in models_or_parameters)
     else:
         counts = sum(v.numel() for v in models_or_parameters)
-    if unit.lower() == "mb":
-        counts /= 1e6
-    elif unit.lower() == "kb":
-        counts /= 1e3
-    elif unit.lower() == "gb":
-        counts /= 1e9
+    unit = unit.lower()
+    if unit == "kb" or unit == "k":
+        counts /= 2 ** 10
+    elif unit == "mb" or unit == "m":
+        counts /= 2 ** 20
+    elif unit == "gb" or unit == "g":
+        counts /= 2 ** 30
     elif unit is not None:
         raise ValueError("Unknow unit: {:}".format(unit))
     return counts
--- a/qlib/contrib/model/pytorch_utils.py
+++ b/qlib/contrib/model/pytorch_utils.py
@@ -1,15 +1,18 @@
 # Copyright (c) Microsoft Corporation.
 # Licensed under the MIT License.
 
-import numpy as np
 import torch.nn as nn
 
 
-def count_parameters(model_or_parameters, unit="mb"):
-    if isinstance(model_or_parameters, nn.Module):
-        counts = np.sum(np.prod(v.size()) for v in model_or_parameters.parameters())
+def count_parameters(models_or_parameters, unit="mb"):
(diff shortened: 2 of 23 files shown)


qlib/workflow/expm.py
```diff
@@ -33,6 +33,10 @@ def __repr__(self):
             name=self.__class__.__name__, duri=self._default_uri, curi=self._current_uri
         )
 
+    def reset_default_uri(self, uri: Text):
+        self._default_uri = uri
```
Derek-Wds, line 37: Is any reason that you want to add this method? From my perspective, `default_uri` should be initialized and fixed when users call `qlib.init`. It seems there is no need to have such method here.

qlib/workflow/recorder.py
```diff
@@ -220,14 +220,45 @@ def __init__(self, experiment_id, uri, name=None, mlflow_run=None):
...
+    @property
+    def uri(self):
+        return self._uri
+
+    @property
+    def artifact_uri(self):
+        return self._artifact_uri
+
+    @property
+    def root_uri(self):
```
Derek-Wds, line 244: What is this method for? Can you give us more explanation about it?

qlib/workflow/__init__.py
```diff
@@ -282,6 +285,12 @@ def get_uri(self):
         """
         return self.exp_manager.uri
 
+    def reset_default_uri(self, uri: Text):
```
Derek-Wds, line 288: Same as the one in `expm`.

qlib/contrib/model/pytorch_utils.py
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+import numpy as np
+import torch.nn as nn
+
+
+def count_parameters(model_or_parameters, unit="mb"):
```
Derek-Wds, line 8: From my perspective, it is not necessary to make this feature as a single function, which actually can be implemented in one line. Here are some [resources](https://pytorch.org/docs/stable/generated/torch.numel.html) and references ([link1](https://discuss.pytorch.org/t/how-do-i-check-the-number-of-parameters-of-a-model/4325/25), [link2](https://stackoverflow.com/questions/49201236/check-the-total-number-of-parameters-in-a-pytorch-model)) may be useful.

If you really want to make it a decent feature, I suggest you can print the number of parameters in a nicer way (for example, similar to `Keras`'s stdout). But this should be carefully design, otherwise it will make the output very messy.

qlib/log.py
```diff
@@ -112,6 +108,27 @@ def set_log_with_config(log_config: dict):
     logging_config.dictConfig(log_config)
 
 
+def set_log_basic_config(filename: Optional[Text] = None, format: Optional[Text] = None, level: Optional[int] = None):
```
Derek-Wds, line 111: Most of the loggers used in `Qlib` are initialized through the function `get_module_logger`. If you want to do some enhancement about setting the `logger`, you may want to make some modifications upon that function instead of creating a new one. This function seems to be a little bit redundant.

qlib/workflow/expm.py
D-X-Y: A simple case is:
after we run many different experiments (with different uri), we may want to compare their performance. To query the metrics of each of these experiments, we need to let R know their uri.

qlib/workflow/recorder.py
D-X-Y: This is to return the path of a specific record. For example, you can call `R.get_recorder().root_uri` and it will return: `ml-runs/2/32792162f0f44951b1285c0b2e489252`.

qlib/workflow/__init__.py
D-X-Y: Yeap, to provide an api to reset default uri. Use to query experiments with different uri.

qlib/log.py
D-X-Y: The motivation is that `get_module_logger` is called in different separate Python files, and it is hard for users to control the detailed behavior of `get_module_logger` function in each file.
This function provides a global control for logging, a case is that before each experiment runs, we can use this function to save the following logs into a file under the recorder directory (without caring about the kwargs of get_module_logger in each file).

qlib/workflow/recorder.py
Derek-Wds: Please update the related `docs`.

qlib/workflow/recorder.py
```diff
@@ -220,14 +220,45 @@ def __init__(self, experiment_id, uri, name=None, mlflow_run=None):
...
+        start_str = "file:"
+        if self.artifact_uri is not None:
+            xpath = self.artifact_uri.strip(start_str)
```
you-n-g, line 247: lstrip will be more accurate

qlib/workflow/recorder.py
you-n-g, line 244: If the funtion aims to get the path in a filesystem, it should raise error(e.g. NotImplementError) when the uri is not file system.

Besides, the `root_uri` is not a good name for user.

D-X-Y:
@you-n-g do you have any suggested name for root_uri?

D-X-Y:
Following our offline discussion, I have:
- temporarily removed set_log functions
- simplify the logic of default uri in expm
- rename root_uri to get_local_dir with more robust logic and doc
- update count_parameters and add doc

Please let me know if you have any other concerns about the current version.

qlib/contrib/backtest/account.py
```diff
@@ -104,10 +104,9 @@ def update_daily_end(self, today, trader):
             # if suspend, no new price to be updated, profit is 0
             if trader.check_stock_suspended(code, today):
                 continue
-            else:
```
Derek-Wds, line 107: Why change this?

D-X-Y: The results are the same while logic is simpler and reduced one line of code.

Derek-Wds (approved):
LGTM. Thanks~

D-X-Y:
@Derek-Wds Thanks for your prompt response.

PR #325, opened 2021-03-10
Tabnet bug fixed

javaThonc:
Fix Tabnet's bug that it will not be stopped at the right epoch and adding alpha360 example of tabnet

### DIFF
--- a/examples/highfreq/high_freq_tree.ipynb
+++ b/examples/highfreq/high_freq_tree.ipynb
@@ -1,0 +1,192 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#  Copyright (c) Microsoft Corporation.\n",
+    "#  Licensed under the MIT License.\n",
+    "import qlib\n",
+    "import os\n",
+    "import random\n",
+    "import pandas as pd\n",
+    "import numpy as np\n",
+    "from multiprocessing import Pool\n",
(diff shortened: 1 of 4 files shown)


examples/highfreq/high_freq_tree.ipynb
```diff
@@ -0,0 +1,192 @@
+{
```
you-n-g, line 1: Why is this file necessary?

examples/highfreq/highfreq_gdbt_model.py
```diff
@@ -0,0 +1,157 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
```
you-n-g, line 2: Why is this file necessary?

you-n-g (commented):
@javaThonc 
Could you update Tabnet's results in the benchmark table?

D-X-Y:
It seems that TabNet does not use the best-validation-performance model for evaluation?

examples/highfreq/high_freq_tree.ipynb
javaThonc: This file is a demo for high-frequency workflow to run in Jupiter notebook and I think it might be straightforward for others to have a python demo.

javaThonc: Since it is really similar to the current workflow example and I will upload a YAML version of it later,   we can also delete it to make the example more concise, depending on whether it is necessary to have a high-frequency version of workflow.

javaThonc:
Close the pr since there is overlapping between prs

PR #329, opened 2021-03-12
Fix Various Bugs for contrib.pytorch_ models

D-X-Y:
There are some potential bugs in the implementation of PyTorch models in qlib

## Description
- `use_gpu` maybe True even if we did not use GPU
- TabNet used "test set" during pre-training
- TabNet used np.inf instead "-np.inf" for the initial best validation score
- Given these bugs in TabNet, I temporarily remove the results of TabNet in README
- Wrap torch.no_grad for evaluation procedures.
- The old MLP models may skip the validation procedure.

## How Has This Been Tested?

## Types of changes

- [x] Fix bugs


you-n-g:
@D-X-Y  Thanks for your contribution!

PR #337, opened 2021-03-14
Fix bugs in Ghost BN in TabNet and typos in README

D-X-Y:
The original implementation of Ghost BN in TabNet is unsafe. It will crash when the input size is small than the number of ghosts, which happens in `csi100`.

## Description
- Fix the bugs in GBN layer in TabNet
- Add config of TabNet on Alpha360
- Fix typos in README

## How Has This Been Tested?

## Types of changes

- [x] Fix bugs
- [x] Update documentation

### DIFF
--- a/qlib/contrib/model/pytorch_tabnet.py
+++ b/qlib/contrib/model/pytorch_tabnet.py
@@ -405,7 +405,6 @@ class FinetuneModel(nn.Module):
 
 
 class DecoderStep(nn.Module):
-
     def __init__(self, inp_dim, out_dim, shared, n_ind, vbs):
         super().__init__()
         self.fea_tran = FeatureTransformer(inp_dim, out_dim, shared, n_ind, vbs)
@@ -417,7 +416,6 @@ class DecoderStep(nn.Module):
 
 
 class TabNet_Decoder(nn.Module):
-
     def __init__(self, inp_dim, out_dim, n_shared, n_ind, vbs, n_steps):
         """
         TabNet decoder that is used in pre-training
@@ -444,7 +442,6 @@ class TabNet_Decoder(nn.Module):
 
 
 class TabNet(nn.Module):
-
     def __init__(self, inp_dim=6, out_dim=6, n_d=64, n_a=64, n_shared=2, n_ind=2, n_steps=5, relax=1.2, vbs=1024):
         """
         TabNet AKA the original encoder
@@ -559,7 +556,6 @@ class AttentionTransformer(nn.Module):
 
 
 class FeatureTransformer(nn.Module):
-
     def __init__(self, inp_dim, out_dim, shared, n_ind, vbs):
         super().__init__()
         first = True
--- a/qlib/contrib/model/pytorch_tabnet.py
+++ b/qlib/contrib/model/pytorch_tabnet.py
@@ -8,7 +8,6 @@ import numpy as np
 import pandas as pd
 import copy
 from sklearn.metrics import roc_auc_score, mean_squared_error
-import logging
 from ...utils import (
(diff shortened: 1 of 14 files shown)


qlib/contrib/model/pytorch_tabnet.py
```diff
@@ -401,7 +397,7 @@ class FinetuneModel(nn.Module):
     """
 
     def __init__(self, input_dim, output_dim, trained_model):
-        super().__init__()
+        super(FinetuneModel, self).__init__()
```
you-n-g, line 400: This is not necessary in Python3

qlib/contrib/model/pytorch_tabnet.py
```diff
@@ -410,9 +406,9 @@ def forward(self, x, priors):
 
 
 class DecoderStep(nn.Module):
-    def __init__(self, inp_dim, out_dim, shared, n_ind, vbs, device):
-        super().__init__()
-        self.fea_tran = FeatureTransformer(inp_dim, out_dim, shared, n_ind, vbs, device)
+    def __init__(self, inp_dim, out_dim, shared, n_ind, vbs):
+        super(DecoderStep, self).__init__()
```
you-n-g, line 410: This is not necessary in Python3

qlib/contrib/model/pytorch_tabnet.py
D-X-Y: Should we follow PyTorch style for these PyTorch models? From FB's official implementation of different PyTorch models (https://github.com/pytorch/vision/blob/master/torchvision/models/googlenet.py#L73), they all use such style.

qlib/contrib/model/pytorch_tabnet.py
D-X-Y: Should we follow PyTorch style for these PyTorch models? From FB's official implementation of different PyTorch models (https://github.com/pytorch/vision/blob/master/torchvision/models/googlenet.py#L73), they all use such style.

Derek-Wds: It is not necessary in fact. The PyTorch style is only needed for inheritance of `nn.Module` class, while in `Qlib` we inherit from our customized `Model` class.

qlib/contrib/model/pytorch_tabnet.py
Derek-Wds: Same as below.

qlib/contrib/model/pytorch_tabnet.py
D-X-Y: This class is the subclass of `nn.Module` not `qlib.Model`~~

qlib/contrib/model/pytorch_tabnet.py
Derek-Wds: PyTorch wrote like this since it wants to be applicable for both Python 2.x and Python 3.x. [Link](https://stackoverflow.com/questions/61288224/why-not-super-init-model-self-in-pytorch). Since Python 2.x is no longer supported and Qlib is only available in Pyhon 3, this is actually useless.

qlib/contrib/model/pytorch_tabnet.py
Derek-Wds: Same as above.

qlib/contrib/model/pytorch_tabnet.py
D-X-Y: gotit~

Derek-Wds (approved):
LGTM. Thanks~

D-X-Y:
> LGTM. Thanks~

Thanks for your help.

PR #393, opened 2021-04-16
Update qlib logger

Derek-Wds:
## Description

The first commit solves the problem of pickling logger object under certain python version. But, I'm still figuring out how to maintain the original docstring of the logger.

**First trial**: now use `help` function will print all the docstrings of `QlibLogger`, but the docstring hints won't show up while using IDEs such as VSCode. 

## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/log.py
+++ b/qlib/log.py
@@ -17,9 +17,9 @@ class MetaLogger(type):
         super().__init__(name, bases, dic)
 
     def __new__(cls, name, bases, dict):
-        wrapper_dict = type(logging.getLogger("module_name")).__dict__.copy()
+        wrapper_dict = type(logging.getLogger("MetaLogger")).__dict__.copy()
         wrapper_dict.update(dict)
-        wrapper_dict["__doc__"] = logging.getLogger("module_name").__doc__
+        wrapper_dict["__doc__"] = logging.getLogger("MetaLogger").__doc__
         return type.__new__(cls, name, bases, wrapper_dict)
 
     def __call__(cls, *args, **kwargs):
--- a/qlib/log.py
+++ b/qlib/log.py
@@ -11,6 +11,26 @@ from contextlib import contextmanager
 
 from .config import C
 
+class QlibLogger(Logger，meta=):
+    '''
+    Customized logger for Qlib.
+    '''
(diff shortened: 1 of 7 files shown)


qlib/log.py
```diff
@@ -12,7 +12,45 @@
 from .config import C
 
 
-def get_module_logger(module_name, level: Optional[int] = None):
+class MetaLogger(type):
+    def __init__(self, name, bases, dic):
+        super().__init__(name, bases, dic)
+
+    def __new__(cls, name, bases, dict):
+        wrapper_dict = type(logging.getLogger("MetaLogger")).__dict__.copy()
```
you-n-g, line 20: Why not use Logger  directly?

qlib/log.py
```diff
@@ -12,7 +12,45 @@
...
+        wrapper_dict.update(dict)
+        wrapper_dict["__doc__"] = logging.getLogger("MetaLogger").__doc__
```
you-n-g, line 22: Why not use `logging.Logger.__doc__` ?

qlib/log.py
```diff
@@ -12,7 +12,45 @@
...
+        return type.__new__(cls, name, bases, wrapper_dict)
+
+    def __call__(cls, *args, **kwargs):
```
you-n-g, line 25: I think `__call__` is not necessary

qlib/log.py
Derek-Wds: Yes we can use `logging.Logger.__doc__` here, which will have the same results as doing so with an instance.

qlib/log.py
you-n-g: Will it be different by replacing  `type(logging.getLogger("MetaLogger"))` with  `logging.Logger`?
Why should we create a Logger instance  here?

qlib/log.py
you-n-g: Why should we create a Logger instance  here instead of using the class? 
Using the class directly takes fewer steps and the code looks simpler

PR #407, opened 2021-04-29
Fix logger pickling error

Derek-Wds:
## Description

The previous PR [#393](https://github.com/microsoft/qlib/pull/393) did not solve the logger pickling completely. The `QlibLogger` object cannot be saved either. This PR is trying to fix the problem and make everything work.

## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/log.py
+++ b/qlib/log.py
@@ -17,6 +17,7 @@ class MetaLogger(type):
         wrapper_dict = logging.Logger.__dict__.copy()
         wrapper_dict.update(dict)
         wrapper_dict["__doc__"] = logging.Logger.__doc__
+        del wrapper_dict["__reduce__"]  # make Logger object can be pickled
         return type.__new__(cls, name, bases, wrapper_dict)
 
 
@@ -28,6 +29,15 @@ class QlibLogger(metaclass=MetaLogger):
     def __init__(self, module_name):
         self.module_name = module_name
         self.level = 0
+
+    def __getstate__(self):
+        return vars(self)
+
+    def __setstate__(self, state):
+        vars(self).update(state)
+
+    def __reduce__(self):
+        return (QlibLogger, (self.module_name,))
 
     @property
     def logger(self):
--- a/qlib/log.py
+++ b/qlib/log.py
@@ -18,7 +18,6 @@ class MetaLogger(type):
         for key in wrapper_dict:
             if key not in dict and key != "__reduce__":
                 dict[key] = wrapper_dict[key]
-        dict["__doc__"] = logging.Logger.__doc__
         return type.__new__(cls, name, bases, dict)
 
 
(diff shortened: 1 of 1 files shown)


qlib/log.py
```diff
@@ -17,6 +17,7 @@ def __new__(cls, name, bases, dict):
         wrapper_dict = logging.Logger.__dict__.copy()
         wrapper_dict.update(dict)
         wrapper_dict["__doc__"] = logging.Logger.__doc__
+        del wrapper_dict["__reduce__"]  # make Logger object can be pickled
```
you-n-g, line 20: I'm little confused about the current implementation.

If the problem is that copying the pickle related magic methods will result in error when pickling, why don't we just exclude them when copying methods dict.

image

qlib/log.py
```diff
@@ -29,6 +31,15 @@ def __init__(self, module_name):
         self.module_name = module_name
         self.level = 0
 
+    def __getstate__(self):
```
you-n-g, line 34: Why  do you have to implement these three magic functions? 
Aren't they the default behaviors of an Python Object?

you-n-g: Maybe these functions should also be skipped in the  `MetaLogger.__new__` function.

Derek-Wds: It is necessary to add these two methods in the class. These functions are used by pickle to customized how objects are pickled. If we do not include these in the class, the `__getattr__` method will be called while pickling and results in recursion error.
Screenshot from 2021-04-30 09-31-04

Besides the original dict did not have these two methods either.
Screenshot from 2021-04-30 09-30-44

It is necessary to specify these two methods if we want to pickle the class.

qlib/log.py
```diff
@@ -29,6 +31,9 @@ def __init__(self, module_name):
         self.module_name = module_name
         self.level = 0
 
+    def __reduce__(self):
```
you-n-g, line 34: Why is reduce necessary here?

qlib/log.py
```diff
@@ -15,9 +15,11 @@
 class MetaLogger(type):
     def __new__(cls, name, bases, dict):
         wrapper_dict = logging.Logger.__dict__.copy()
-        wrapper_dict.update(dict)
-        wrapper_dict["__doc__"] = logging.Logger.__doc__
-        return type.__new__(cls, name, bases, wrapper_dict)
+        for key in wrapper_dict:
+            if key not in dict and key != "__reduce__":
+                dict[key] = wrapper_dict[key]
+        dict["__doc__"] = logging.Logger.__doc__
```
you-n-g, line 21: image

PR #431, opened 2021-05-18
Fix `get_module_by_module_path` to support pickle module loaded from arbitrage source file

evanzd:
## Description
Currently the following script will fail with pickle dump errors:
```python
"""<model.py>
class Model:
    pass
"""
import pickle
from qlib.utils import get_module_by_module_path

model = get_module_by_module_path('model.py').Model()
print(pickle.dumps(model)[:20])
```
image

In this commit, the loaded module is added to `sys.modules` to support pickling.

## Types of changes
- [X ] Fix bugs

### DIFF
--- a/qlib/utils/__init__.py
+++ b/qlib/utils/__init__.py
@@ -179,7 +179,7 @@ def get_module_by_module_path(module_path: Union[str, ModuleType]):
         module = module_path
     else:
         if module_path.endswith(".py"):
-            module_name = "dummy"
+            module_name = re.sub("^[^a-zA-Z_]+", "", re.sub("[^0-9a-zA-Z_]", "", module_path[:-3].replace("/", "_")))
             module_spec = importlib.util.spec_from_file_location(module_name, module_path)
             module = importlib.util.module_from_spec(module_spec)
             sys.modules[module_name] = module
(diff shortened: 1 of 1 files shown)


qlib/utils/__init__.py
```diff
@@ -178,8 +179,10 @@ def get_module_by_module_path(module_path: Union[str, ModuleType]):
         module = module_path
     else:
         if module_path.endswith(".py"):
-            module_spec = importlib.util.spec_from_file_location("", module_path)
+            module_name = re.sub("^[^a-zA-Z_]+", "", re.sub("[^0-9a-zA-Z_]", "", module_path[:-3].replace("/", "_")))
```
you-n-g, line 182: Why can such behavior solve this problem?
Could you give more docs about it ?

evanzd: I'm referring the official doc to add `module_name` in `sys.modules` to make the loaded module picklable.
https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly

The only thing we need is a unique `module_name` without any possible collision. As module name is also a variable in python, this implementation follows this discussion:
https://stackoverflow.com/questions/3303312/how-do-i-convert-a-string-to-a-valid-variable-name-in-python

you-n-g: @evanzd   Thanks a lot. 
学到了， 学到了!

PR #438, opened 2021-05-25
Support Highfreq Backtest with the Model/Rule/RL Strategy

bxdd:
## Description

Support Highfreq Backtest with the Model/Rule/RL Strategy

## How Has This Been Tested?
- [ ] Pass the test by running: `pytest qlib/tests/test_all_pipeline.py` under upper directory of `qlib`.
- [ ] If you are adding a new feature, test on your own test scripts.

## Screenshots of Test Results (if appropriate):
1. Pipeline test:
2. Your own tests:

## Types of changes

- [ ] Fix bugs
- [x] Add new feature
- [ ] Update documentation

## TODOs
- [ ] release new version of Qlib to fix qrun of  users which install Qlib by `pip install pyqlib`

PR #456, opened 2021-06-01
Dummy RL example on nested decision framework

ultmaster:
## Motivation and Context

See #438.

## Description

This PR provides a demonstration on how to support RL with current nested decision framework.

### Design

- EpisodicState and StepState: data structure that contains information of a trajectory in RL language.
  - Summarize the metrics for now. Will probably be moved into other modules in future.
- Observation, Action, Policy: **stateless** components that functions as translator between policy and environment.
  - `EpisodicState.from_order_and_executor` and `create_sub_order` are designed as the interpreter between RL data structure and `Order`. Currently, data structure and functions in RL handles one order each. Therefore, these two utility function also groups the orders by stock id and direction.
- SingleOrderEnv: Wrapper of all the utility function above, used for RL training.

## Issues

- [ ] Exchange is not robust enough. Invalid sub-orders sometimes cause executor to crash.
  - [ ] How to deal with trade unit, suspended stocks in RL? *Current: do nothing.*
- [ ] Data-multiprocessing has conflict with RL-multiprocessing. Might need refact.
- [ ] Needs tests/data on EP's minute-level data.
  - [x] Data is not ready.
  - [ ] The current system is *super slow*.
- [ ] API (minor) issues
  - [ ] How to get inner frequency and trade len in strategy? *Current: hardcode.*
  - [ ] Same configuration needs to be specified multiple times, which looks weird. e.g., `start_time` in benchmark and backtest_func, `freq` in exchange, executor, and strategy.
  - [ ] Coding experience with `infra` is still not smooth enough. No code-completion/suggestion, multiple ways to access/set infra (member var, init, reset, get method, ...).


bxdd (commented):
`assets/order` is a BIN file, should it be managed by git?

you-n-g:
@ultmaster Please add requirements.txt for your specific example if extra packages are required.

ultmaster:
> @ultmaster Please add requirements.txt for your specific example if extra packages are required.

done

ultmaster:
> `assets/order` is a BIN file, should it be managed by git?

I think so. I put it here only for test purposes. It should be deleted before merging.

ultmaster:
# Future TODOs

- Refine the implementation of RL env and leverage the QlibIntRLEnv in qlib.
- Calculate PA per step in qlib.
-

PR #466, opened 2021-06-11
Online bug fix, enhancement &  docs for dataset, workflow, trainer ...

you-n-g:
## Description
Fix following bugs
- Skip duplicated qlib.auto_init()
- Fix TSDatasetH flt_col bug!
- Resolve qlib log attribute confliction
- Trainer API enhancement
- More docs and user-friendly warning

## How Has This Been Tested?
- [X] Pass the test by running: `pytest qlib/tests/test_all_pipeline.py` under upper directory of `qlib`.

## Types of changes

- [X] Fix bugs
- [X] Add new feature
- [X] Update documentation

### DIFF
--- a/qlib/__init__.py
+++ b/qlib/__init__.py
@@ -20,11 +20,17 @@ def init(default_conf="client", **kwargs):
     from .config import C
     from .data.cache import H
 
-    H.clear()
-
     # FIXME: this logger ignored the level in config
     logger = get_module_logger("Initialization", level=logging.INFO)
 
+    skip_if_reg = kwargs.pop("skip_if_reg", False)
+    if skip_if_reg and C.registered:
+        # if we reinitialize Qlib during running an experiment `R.start`.
+        # it will result in loss of the recorder
+        logger.warning("Skip initialization because `skip_if_reg is True`")
+        return
+
+    H.clear()
     C.set(default_conf, **kwargs)
 
     # check path if server/local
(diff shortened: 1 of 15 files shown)


qlib/__init__.py
```diff
@@ -20,11 +20,17 @@ def init(default_conf="client", **kwargs):
     from .config import C
     from .data.cache import H
 
-    H.clear()
-
     # FIXME: this logger ignored the level in config
     logger = get_module_logger("Initialization", level=logging.INFO)
 
+    skip_if_reg = kwargs.pop("skip_if_reg", False)
+    if skip_if_reg and C.registered:
```
Derek-Wds, line 27: Should we also check if `R` contains active experiment here? It seems that only the function `auto_init` has been enabled skipping initialization for now.

qlib/workflow/expm.py
```diff
@@ -351,6 +351,15 @@ def _get_exp(self, experiment_id=None, experiment_name=None):
             experiment_id is not None or experiment_name is not None
         ), "Please input at least one of experiment/recorder id or name before retrieving experiment/recorder."
         if experiment_id is not None:
+            try:
```
Derek-Wds, line 354: I feel like this is a little bit redundant? If the user inputs an invalid experiment id, then mlflow will not find the corresponding experiment and we will raise an `MlflowException` in the following codes (under the `except` [here](https://github.com/microsoft/qlib/blob/d4b36bdab448859f04b163b89843594660b3323f/qlib/workflow/expm.py#L369)).

Instead, we could update the `MlflowException` error messages, which tells the user that the id of an experiment should be `int` or `str` of int.

Derek-Wds: Maybe we could try to add `*` in the method to force users to specify the correct arguments when calling `get_exp`.

Derek-Wds (approved):
LGTM! :heart:

PR #478, opened 2021-06-24
Set self.fitted = True instead of self._fitted.

topskychen:
All the current models are using fitted to indicate the model is fitted. Current `_fitted` should be a bug due to legacy code.

## Description

Use `self.fitted = True` instead of `self._fitted = True` in pytorch_nn.py.

## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation


you-n-g:
@topskychen  Thanks so much for fixing the bug!

PR #484, opened 2021-06-29
Update initialization.rst

xiaowuhu:
## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [ ] Fix bugs
- [ ] Add new feature
- [x] Update documentation

### DIFF
--- a/docs/start/initialization.rst
+++ b/docs/start/initialization.rst
@@ -12,12 +12,20 @@ Initialization
 
 Please follow the steps below to initialize ``Qlib``.
 
+If you use "pip install pyqlib" in the installation, you need github clone from https://github.com/microsoft/qlib to you local disk, then you will get all the source code. Feel free to use it, change it, or even send your PR to us.
+
 Download and prepare the Data: execute the following command to download stock data. Please pay `attention` that the data is collected from `Yahoo Finance <https://finance.yahoo.com/lookup>`_ and the data might not be perfect. We recommend users to prepare their own data if they have high-quality datasets. Please refer to `Data <../component/data.html#converting-csv-format-into-qlib-format>`_ for more information about customized dataset.
     
     .. code-block:: bash
     
         python scripts/get_data.py qlib_data --target_dir ~/.qlib/qlib_data/cn_data --region cn
-        
+
+if you are using Windows OS, first create a direction such as "md E:/Qlib/qlib_data", then using below command to get data (assume you clone qlib source code under e:/github/qlib):
+
+    .. code-block:: bash
+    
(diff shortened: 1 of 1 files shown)


docs/start/initialization.rst
```diff
@@ -12,12 +12,20 @@ Initialization
...
     .. code-block:: bash
     
         python scripts/get_data.py qlib_data --target_dir ~/.qlib/qlib_data/cn_data --region cn
-        
+
+if you are using Windows OS, first create a directory such as "md E:/Qlib/qlib_data", then using below command to get data (assume you clone qlib source code under e:/github/qlib):
+
+    .. code-block:: bash
+    
+        python e:/github/qlib/scripts/get_data.py qlib_data --target_dir e:/Qlib/qlib_data/cn_data --region cn
```
you-n-g, line 27: Qlib use  [~/.qlib/qlib_data/cn_data](https://github.com/xiaowuhu/qlib/blob/patch-1/qlib/config.py#L183) as the default data source. 
Using it as the target will be more user-friendly.

Another option is ignoring this  parameter to fallback to the [default value](https://github.com/lzh222333/qlib/blob/qlib/main/qlib/tests/data.py#L111)

xiaowuhu: in Windows OS, we cannot use "~/.qlib/......" as directory name

you-n-g: image
I think we can use this variable.

Another option is ignoring this parameter to fallback to the [default value](https://github.com/lzh222333/qlib/blob/qlib/main/qlib/tests/data.py#L111)

zhupr: Using `pathlib.Path().expanduser()`: https://github.com/microsoft/qlib/blob/main/qlib/tests/data.py#L42 (**python version >= 3.4**)
image

xiaowuhu: This works. 但是，通常用户会考虑到用一个大容量的驱动器来装载数据，而不是放在user\{alias}下面或者Home目录下(因为它会占用C盘的空间，而通常用户会考虑把数据放到D盘上，这样以后万一重新安装操作系统时，可以保留数据）。所以我个人认为用绝对路径是一个best practice.

you-n-g: The different settings will make subsequent steps fail.
For example:
https://qlib.readthedocs.io/en/latest/start/initialization.html#id1

I think we can tell users that the `uri`is configurable in the docs and comments.

For tutorial purposes,  running successfully is more important than robustness.

PR #512, opened 2021-07-14
Add filter string to mlflow experiment list_recorders function.

chaosddp:
## Description

Current RecorderCollector use default list_recorders function list all ACTIVE_ONLY runs, it may cost much time if there are too much active runs, with mlflow filter string support, we can reduce this number.

## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [ ] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/workflow/task/collect.py
+++ b/qlib/workflow/task/collect.py
@@ -139,6 +139,7 @@ class RecorderCollector(Collector):
         rec_filter_func=None,
         artifacts_path={"pred": "pred.pkl"},
         artifacts_key=None,
+        filter_string: str = ""
     ):
         """
         Init RecorderCollector.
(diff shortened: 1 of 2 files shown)


qlib/workflow/task/collect.py
```diff
@@ -139,6 +139,7 @@ def __init__(
         rec_filter_func=None,
         artifacts_path={"pred": "pred.pkl"},
         artifacts_key=None,
+        filter_string: str = "",
```
you-n-g, line 142: Qlib's framework is decoupled with mlflow.
So these filter_string specific things shouldn't appear here.
I think `list_kwargs={}`  and `recs = self.experiment.list_recorders(**self.list_kwargs)` will make it more  loosely-coupled

chaosddp: fixed as recommend :)

you-n-g:
@chaosddp 
It looks great now! Thanks :)

PR #600, opened 2021-09-14
Update setup.py

anukaal:
updated classifiers

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, macos-latest]
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8, 3.9]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


Wangwuyi123:
We tried to create a new 3.9 environment by opening qlib/examples/workflow_by_code.ipynb test jupyter，and there was no problem. 

If your test results are normal, please change qlib/examples/workflow_by_code.ipynb line16 to python-version: [3.6, 3.7, 3.8,3.9].

Wangwuyi123:
Change matplotlib in setup.py to matplotlib==3.3

First test whether qlib supports python3.9. The test method is to run workflow_by_code.ipynb. If the result is normal, it means that python3.9 is supported.

Currently qlib does not support packaging 3.9 version to pypi, you may need to modify this place: qlib/python-publish.yml at main · microsoft/qlib (github.com)

Wangwuyi123:
Change matplotlib in setup.py to `matplotlib==3.3`

Test whether qlib supports `python 3.9`, by running `workflow_by_code.ipynb`, if the result is normal, it means that python 3.9 is supported.

Currently qlib does not support packing 3.9 to pypi, you may need to change this place https://github.com/microsoft/qlib/blob/main/.github/workflows/python-publish.yml#L16

Issue #731, opened 2021-12-08
install from source with pip install . will cause import error: FileNotFoundError: [Errno 2] No such file or directory: '~/qlib/qlib/VERSION.txt'


b4thesunrise:
## To Reproduce

1.install the latest master branch with pip install .
2.python, import qlib
3. FileNotFoundError: [Errno 2] No such file or directory: '~/qlib/qlib/VERSION.txt'

## Screenshot

![image](https://user-images.githubusercontent.com/43727375/145143163-15b00bd2-18d9-4946-adb5-77dd074ce182.png)

## Environment

**Note**: User could run `cd scripts && python collect_info.py all` under project directory to get system information
and paste them here directly.

 - Qlib version:0.8.0.99
 - Python version:3.7
 - OS (`Windows`, `Linux`, `MacOS`):Macos
 - Commit number (optional, please provide it if you are using the dev version):

Wangwuyi123:
@b4thesunrise 
According to your environment, I did not reproduce the same bug, please test it in your environment again, if the same bug reproduces, please be sure to inform.

you-n-g:
@b4thesunrise 
Thanks so much  for your contribution.

The  VERSION.txt will be automatically created when you run `python setup.py install` or `python setup.py develop` 
https://github.com/microsoft/qlib/blob/main/setup.py#L19

Did you encounter any problems when creating it?

b4thesunrise:
> @b4thesunrise Thanks so much for your contribution.
> 
> The VERSION.txt will be automatically created when you run `python setup.py install` or `python setup.py develop` https://github.com/microsoft/qlib/blob/main/setup.py#L19
> 
> Did you encounter any problems when creating it?

> @b4thesunrise According to your environment, I did not reproduce the same bug, please test it in your environment again, if the same bug reproduces, please be sure to inform.

I haven't encounter problem when creating the version.txt file by `python setup.py install`, the reason why I encounter the problem is I first use `python setup.py install` to  install qlib at first, after that, I tried to update the version by git clone, cd qlib, and `python install .`, and encounter the FileNotFoundError. I think that's because I understand the documentation wrongly, thanks for your help!

you-n-g:
@b4thesunrise 
I tried `pip install .` later and encountered the same error. 
It is indeed a bug that should be fixed.
Thanks for your reporting

Looking forward to your update on [this pr](https://github.com/microsoft/qlib/pull/732)

PR #732, opened 2021-12-08
solve VERSION.txt bug

b4thesunrise:
## Description

change .gitignore file
add back version,txt
## Motivation and Context

https://github.com/microsoft/qlib/issues/731

solve the problem of install qlib with pip install .
## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [ ] Fix bugs
- [ ] Add new feature
- [ ] Update documentation


zhupr:
@b4thesunrise Hi,
When installing from source (`python setup.py develop/install`) will automatically copy `src/VERSION.txt` to `src/qlib`: https://github.com/microsoft/qlib/blob/main/setup.py#L20

zhupr:
@b4thesunrise Hi,
`pip install <source>` will not execute the `copyfile` in `setup.py`.
Can you help to change the "version" of `qlib` to something like "pip", refer to: https://github.com/pypa/pip/blob/main/setup.py#L28

b4thesunrise:
> @b4thesunrise Hi, `pip install <source>` will not execute the `copyfile` in `setup.py`. Can you help to change the "version" of `qlib` to something like "pip", refer to: https://github.com/pypa/pip/blob/main/setup.py#L28

Hi! I am willing to offer help, so the problem is `pip install <source>`  will cause the error of the lack of version.txt since the copy, so we want to get the version with function like `get version` in pip instead of use `copyfile`?

zhupr:
> > @b4thesunrise Hi, `pip install <source>` will not execute the `copyfile` in `setup.py`. Can you help to change the "version" of `qlib` to something like "pip", refer to: https://github.com/pypa/pip/blob/main/setup.py#L28
> 
> Hi! I am willing to offer help, so the problem is `pip install <source>` will cause the error of the lack of version.txt since the copy, so we want to get the version with function like `get version` in pip instead of use `copyfile`?

Yes, you need to implement a `get_version` function to get the version from `qlib/__init__.py`

b4thesunrise:
> > > @b4thesunrise Hi, `pip install <source>` will not execute the `copyfile` in `setup.py`. Can you help to change the "version" of `qlib` to something like "pip", refer to: https://github.com/pypa/pip/blob/main/setup.py#L28
> > 
> > 
> > Hi! I am willing to offer help, so the problem is `pip install <source>` will cause the error of the lack of version.txt since the copy, so we want to get the version with function like `get version` in pip instead of use `copyfile`?
> 
> Yes, you need to implement a `get_version` function to get the version from `qlib/__init__.py`

got it, will work on it

b4thesunrise:
have already followed the style of pypi and use black format, please give a review :)

you-n-g:
@b4thesunrise  It looks great! 
Thanks so much for your contribution.
Welcome to become Qlib's [contributor](https://github.com/microsoft/qlib/graphs/contributors)!!

b4thesunrise:
> @b4thesunrise It looks great! Thanks so much for your contribution. Welcome to become Qlib's [contributor](https://github.com/microsoft/qlib/graphs/contributors)!!

Thanks! My pleasure!

PR #781, opened 2021-12-27
Reformat example data names: use `{region}_data` for 1-day data, and `{region}_data_1min` for 1-min data

cning112:
## Description
Use the same name for 1min data

## Motivation and Context
The same data has two names, meaning the same data needs to be downloaded twice to run some examples.

## How Has This Been Tested?
- not needed
- 

1. Pipeline test:
2. Your own tests:

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,10 +151,10 @@ Also, users can install the latest dev version ``Qlib`` by the source code accor
 Load and prepare data by running the following code:
   ```bash
   # get 1d data
-  python scripts/get_data.py qlib_data --target_dir ~/.qlib/qlib_data/cn_data --region cn
+  python scripts/get_data.py qlib_data --target_dir ~/.qlib/qlib_data/qlib_cn_1d --region cn
 
   # get 1min data
-  python scripts/get_data.py qlib_data --target_dir ~/.qlib/qlib_data/cn_data_1min --region cn --interval 1min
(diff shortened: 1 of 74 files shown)


you-n-g:
@cning112 Thanks for your contribution.

I found a lot of commands with different format
image

Would you like to format them to the same path? 
For example, `qlib_cn_1min` is a good name.

you-n-g:
Besides ...
image

Users from the community will be grateful to you :P

cning112:
sure, will make the change in the next commit. Thanks

cning112:
Actually I think the `qlib_` prefix can be removed from the example data names. `{region}_{freq}` should suffice. What do you think? @you-n-g

you-n-g:
@cning112  Yes. Removing the redundant prefix make it simpler
Besides making it simpler, it will be better to make the name consistent with the previous directory to be compatible.
I think keeping "~/.qlib/qlib_data/cn_data" for 1-day data will be more friendly to old users

you-n-g:
@cning112   Please check the Paths in CI workflows
https://github.com/microsoft/qlib/tree/main/.github/workflows

Thanks

you-n-g:
It looks great! Thanks so much!

Issue #840, opened 2022-01-10
Add test cases for operators.


ChiahungTai:
## Motivation

1. Application scenario
2. Related works (Papers, Github repos etc.):
3. Any other relevant and important information:

Add test cases for operators.
Test case is very important for code quality.

ChiahungTai:
Merged.

PR #841, opened 2022-01-10
[840] - Test case for operators.

ChiahungTai:
## Description

This is a draft codes for the operators test cases.

## Motivation and Context

More test cases, better code quality
#840 

## How Has This Been Tested?
- [x] Pass the test by running: `pytest qlib/tests/ops/test_elem_operator.py.py` under upper directory of `qlib`.
N/A

1. Pipeline test:
2. Your own tests:
Screen Shot 2022-01-10 at 9 39 46 PM

## Types of changes

- [ ] Fix bugs
- [x] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/tests/__init__.py
+++ b/qlib/tests/__init__.py
@@ -2,6 +2,9 @@ import unittest
 from .data import GetData
 from .. import init
 from ..constant import REG_CN
+from qlib.data.filter import NameDFilter
+from qlib.data import D
+from qlib.data.data import Cal, DatasetD
 
 
 class TestAutoData(unittest.TestCase):
@@ -56,10 +59,6 @@ class TestAutoData(unittest.TestCase):
 class TestOperatorData(TestAutoData):
     @classmethod
     def setUpClass(cls, enable_1d_type="simple", enable_1min=False) -> None:
-        from qlib.data.filter import NameDFilter
-        from qlib.data import D
-        from qlib.data.data import Cal, DatasetD
-
         # use default data
         super().setUpClass(enable_1d_type, enable_1min)
         nameDFilter = NameDFilter(name_rule_re="SH600110")
@@ -76,7 +75,3 @@ class TestOperatorData(TestAutoData):
         cls.end_time = cal[-1]
         cls.inst = list(instruments_d.keys())[0]
         cls.spans = list(instruments_d.values())[0]
-
-    def setUp(self) -> None:
-        self.assertEqual(len(self.instruments_d), 1)
-        self.assertGreater(len(self.cal), 0)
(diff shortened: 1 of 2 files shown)


ChiahungTai:
@you-n-g Could you please give me some feedbacks for this patch?
Thank you.

qlib/tests/__init__.py
```diff
@@ -51,3 +51,32 @@ def setUpClass(cls, enable_1d_type="simple", enable_1min=False) -> None:
             dataset_cache=None,
             **cls._setup_kwargs,
         )
+
+
+class TestOperatorData(TestAutoData):
+    @classmethod
+    def setUpClass(cls, enable_1d_type="simple", enable_1min=False) -> None:
+        from qlib.data.filter import NameDFilter
```
you-n-g, line 59: Why do you import them in the method instead of at the head of this file?

qlib/tests/__init__.py
```diff
@@ -51,3 +51,32 @@ def setUpClass(cls, enable_1d_type="simple", enable_1min=False) -> None:
...
+        instruments_d = DatasetD.get_instruments_d(instruments, freq)
+        cls.instruments_d = instruments_d
+        cal = Cal.calendar(start_time, end_time, freq)
+        cls.cal = cal
+        cls.start_time = cal[0]
+        cls.end_time = cal[-1]
+        cls.inst = list(instruments_d.keys())[0]
+        cls.spans = list(instruments_d.values())[0]
+
+    def setUp(self) -> None:
```
you-n-g, line 80: It looks a little weird here.
Users expect `setUp` to set up sth.  But it actually tests sth.

Maybe you can extract this method into an individual class like this instead of putting it in the base class.

```python
class TestOperatorDataSetting(TestOperatorData):
      def test_setting(self):
            self.assertEqual(len(self.instruments_d), 1)
            self.assertGreater(len(self.cal), 0)
```

qlib/tests/__init__.py
ChiahungTai: Put import to the head of file is good to me. 

I mimic from here.
https://github.com/microsoft/qlib/blob/19689024d49e61e1187990d2c5f35b6e45ec513c/qlib/backtest/__init__.py#L174
Why would you guys import in the method in above case.
Thanks.

qlib/tests/__init__.py
ChiahungTai: No problem, will do.

qlib/tests/__init__.py
you-n-g: That is for avoiding recursive import.
Both of the two modules import each other.

ChiahungTai: Got it.

you-n-g:
LGTM! Thanks so much!

PR #959, opened 2022-03-10
feat: add instrument context to inst_processor

PalanQu:
Add instrument context to inst_processor
## Description

Many indicator analysis need to use forward adjust stock price, but this price need to be calculated when load the data, so I need to write inst_processor to pre-calculate the forward adjust stock price like following, I want to read the latest adjfactor from source data, but there is no context in __call__.
``` python
class AdjPriceProcessor(InstProcessor):
    def __call__(self, df: pd.DataFrame, *args, **kwargs):
        # TODO: Read the latest adjfactor from source data
        return df

day_fields = ['$close']
day_names = ['close']
labels = ['Ref($close, -2)/Ref($close, -1) - 1'] # label
label_names = ['LABEL']
def feature_config():
    return {
        'feature_day': (
            day_fields,
            day_names,
        ),
        'label': (labels, label_names)
    }

data_loader = QlibDataLoader(
    config=feature_config(),
    freq={
        'label': 'day',
        'feature_day': 'day',
    },
    inst_processor={
        'feature_day': [AdjPriceProcessor()],
        'label': [],
    }
)

df = data_loader.load(instruments=['SH600519', 'SZ300750'], start_time='2020-01-01', end_time='2021-01-01')
```

## How Has This Been Tested?
<! ---  Put an `x` in all the boxes that apply: --->
- [x ] Pass the test by running: `pytest qlib/tests/test_all_pipeline.py` under upper directory of `qlib`.

1. Pipeline test:
2. Your own tests:

## Types of changes

- [ ] Fix bugs
- [ x] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/data/data.py
+++ b/qlib/data/data.py
@@ -617,7 +617,7 @@ class DatasetProvider(abc.ABC):
         for _processor in inst_processors:
             if _processor:
                 _processor_obj = init_instance_by_config(_processor, accept_types=InstProcessor)
-                data = _processor_obj(data)
+                data = _processor_obj(data, instrument=inst)
         return data
 
 
--- a/qlib/data/inst_processor.py
+++ b/qlib/data/inst_processor.py
@@ -5,7 +5,7 @@ import pandas as pd
 
 class InstProcessor:
     @abc.abstractmethod
-    def __call__(self, df: pd.DataFrame, *args, **kwargs):
+    def __call__(self, df: pd.DataFrame, instrument, *args, **kwargs):
         """
         process the data
 


qlib/data/data.py
```diff
@@ -617,7 +617,7 @@ def inst_calculator(inst, start_time, end_time, freq, column_names, spans=None,
         for _processor in inst_processors:
             if _processor:
                 _processor_obj = init_instance_by_config(_processor, accept_types=InstProcessor)
-                data = _processor_obj(data)
+                data = _processor_obj(data, instrument=inst)
```
you-n-g, line 620: Thanks for your advice!  It looks more reasonable to add this parameter.

In addition, I think the interface should be changed with it
https://github.com/microsoft/qlib/blob/main/qlib/data/inst_processor.py#L8

you-n-g:
LGTM! Thanks !

PR #990, opened 2022-03-17
Ibovespa index support

### LINKED ISSUE #956: Add new dataset and index: ibovespa and B3 datasets from Brazil Stock Exchange
## 🌟 Feature Description

Brazil's stock exchange is one of the biggest in the world and therefore adding its dataset into qlib can help people do research via qlib infrastructure using different datasets.

## Motivation

1. Application scenario: support brazilian stocks exchange investment strategy research.
2. Related works (Papers, Github repos etc.): Not applicable
3. Any other relevant and important information: More details on Brazil's stock exchange can be found on [B3](https://www.b3.com.br/en_us/)

## Alternatives

No other alternative have been considered, since the data is required to proceed with an investment strategy research.

## Additional Notes

The ideia is to 
- add code into `data_collector` to get data about the largest index in Brazil's stocks exchange - ibovespa
- add code into `yahoo` folder in order to download stocks historic prices.

igor17400:
## Description

- For the **feat: download ibovespa index historic composition**  the following files were included:
  - `scripts/data_collector/br_index/README.md`
  - `scripts/data_collector/br_index/collector.py`
  - `scripts/data_collector/br_index/requirements.txt`
- For **fix: typo error instead of end_date, it was written end_ate** the following files were modified:
  - `scripts/data_collector/index.py`
- For **feat: adds support for downloading stocks historic prices from Brazil's stocks exchange (B3)** the following files were modified:
  - `scripts/data_collector/utils.py`
  - `scripts/data_collector/yahoo/README.md`
  - `scripts/data_collector/yahoo/collector.py`

## Motivation and Context

Brazil's stock exchange is one of the biggest in the world, therefore adding its dataset into qlib can help people do research via qlib infrastructure using different datasets.

1. Application scenario: support brazilian stocks exchange investment strategy research.
2. Related works (Papers, Github repos etc.): Not applicable
3. Any other relevant and important information: More details on Brazil's stock exchange can be found on [B3](https://www.b3.com.br/en_us/)

## How Has This Been Tested?

There isn't a specific script for test, however the added code can be tested running the following commands:

- `python scripts/data_collector/br_index/collector.py --index_name IBOV --qlib_dir ~/.qlib/qlib_data/br_data --method parse_instruments`
- `python scripts/data_collector/br_index/collector.py --index_name IBOV --qlib_dir ~/.qlib/qlib_data/br_data --method save_new_companies`
- `python scripts/data_collector/br_index/collector.py download_data --source_dir ~/.qlib/stock_data/source/br_data --start 2003-01-03 --end 2022-03-01 --delay 1 --interval 1d --region BR`
- `python scripts/data_collector/br_index/collector.py download_data --source_dir ~/.qlib/stock_data/source/br_data_1min --delay 1 --interval 1min --region BR`

1. Pipeline test: Not applicable
2. Your own tests: Not applicable

## Types of changes

- [x] Add new feature
- [x] Update documentation (README.md)

closes #956


ChiahungTai:
@you-n-g Should it rerun the mac os test?

scripts/data_collector/br_index/collector.py
```diff
@@ -0,0 +1,317 @@
...
+    "2019_1Q",
+    "2019_2Q",
+    "2019_3Q",
+    "2020_1Q",
+    "2020_2Q",
+    "2020_3Q",
+    "2021_1Q",
+    "2021_2Q",
+    "2021_3Q",
+    "2022_1Q",
```
you-n-g, line 78: Hi,
Why don't use a for loop to generate this list  until  `datetime.datetime.today()`

scripts/data_collector/br_index/collector.py
```diff
@@ -0,0 +1,317 @@
...
+from loguru import logger
+
+CUR_DIR = Path(__file__).resolve().parent
+sys.path.append(str(CUR_DIR.parent.parent))
+
+from data_collector.index import IndexBase
+
+IBOV_INDEX_COMPOSITION = "https://raw.githubusercontent.com/igor17400/IBOV-HCI/main/historic_composition/{}.csv"
+
+YEAR_QUARTER = [
```
you-n-g, line 20: I think this will be more reasonable to make it as an class attribute  of IBOVIndex

scripts/data_collector/utils.py
```diff
@@ -19,6 +19,7 @@
 from tqdm import tqdm
 from functools import partial
 from concurrent.futures import ProcessPoolExecutor
+from bs4 import BeautifulSoup
```
you-n-g, line 22: Maybe BeautifulSoup should be in the `requirements.txt` ?

SunsetWolf:
In the course of reviewing this PR we found two issues.
1. there are multiple places where the get_instruments() method is used, and we feel that scripts.index.py is the best place for the get_instruments() method to go.
2. data_collector.utils has some very generic stuff put inside it.
Are you willing to fix them? If not, I will fix these issues in the future.

scripts/data_collector/br_index/collector.py
igor17400: Great!
Thanks for the suggestion
I'll work on this change

scripts/data_collector/br_index/collector.py
igor17400: Ook

igor17400:
@SunsetWolf I can work on this!
I'll give it a look

igor17400:
I was checking some functions in the Yahoo `collector.py` code and I suspect I may have found another issue.


> The callable method returns True if the object passed appears callable. If not, it returns False. As the image below shows,

Screen Shot 2022-03-31 at 19 05 43

I was trying to set the `retry` parameter from the decorator `deco_retry` to 1. Because, Yahoo finance does not keeps track of the majority of Brazilian stocks and as a consequence the code is taking a long time to finish. And I was doing so by calling the decorator as `@deco_retry(retry=1)` and defining `retry` variable inside the `_get_ibovespa()` function as the code snipet below shows.

However, it seems that the code below for the decorator `deco_retry` will never return `True` for `callable(retry)` since `retry` variable is an `int` which is not callable by its own nature.

**_Am I doing something wrong?_**

```
def deco_retry(retry: int = 5, retry_sleep: int = 3):
    def deco_func(func):
        @functools.wraps(func)
        def wrapper(*args, **kwargs):
            _retry = 5 if callable(retry) else retry
            _result = None
            for _i in range(1, _retry + 1):
                try:
                    _result = func(*args, **kwargs)
                    break

                except Exception as e:
                    logger.warning(f"{func.__name__}: {_i} :{e}")
                    if _i == _retry:
                        raise

                time.sleep(retry_sleep)
            return _result

        return wrapper

    return deco_func(retry) if callable(retry) else deco_func
```

```
 @deco_retry(retry=1)
    def _get_ibovespa():
        retry = 1
        _symbols = []
        url = "https://www.fundamentus.com.br/detalhes.php?papel="

        # Request
        agent = {"User-Agent": "Mozilla/5.0"}
        page = requests.get(url, headers=agent)

        # BeautifulSoup
        soup = BeautifulSoup(page.content, "html.parser")
        tbody = soup.find("tbody")

        children = tbody.findChildren("a", recursive=True)
        for child in children:
            _symbols.append(str(child).split('"')[-1].split(">")[1].split("<")[0])

        return _symbols
```

scripts/data_collector/us_index/collector.py
```diff
@@ -269,46 +270,6 @@ def parse_instruments(self):
...
-    """
-    _cur_module = importlib.import_module("data_collector.us_index.collector")
-    obj = getattr(_cur_module, f"{index_name.upper()}Index")(
-        qlib_dir=qlib_dir, index_name=index_name, freq=freq, request_retry=request_retry, retry_sleep=retry_sleep
-    )
-    getattr(obj, method)()
-
 
 if __name__ == "__main__":
     fire.Fire(get_instruments)
```
you-n-g, line 275:
```suggestion
    fire.Fire(partial(get_instruments, market_index="br_index" ))
```
- This applies to other index data source as well.


Do you think this will make the interface easier?
Then all the  collector.py CLI  in each  folder can remove a redundant arguments.

scripts/data_collector/br_index/README.md
```diff
@@ -53,9 +53,9 @@ With that reference, the index's composition can be compared quarter by quarter
 
 ```bash
 # parse instruments, using in qlib/instruments.
-python collector.py --index_name IBOV --qlib_dir ~/.qlib/qlib_data/br_data --method parse_instruments
+python collector.py --index_name IBOV --qlib_dir ~/.qlib/qlib_data/br_data --method parse_instruments --market_index br_index
```
you-n-g, line 56: image

scripts/data_collector/utils.py
```diff
@@ -564,3 +562,47 @@ def generate_minutes_calendar_from_daily(
 
 if __name__ == "__main__":
```
you-n-g, line 563: It will be better if `ifmain` is at the bottom of the script.

scripts/data_collector/yahoo/collector.py
```diff
@@ -147,7 +148,29 @@ def _show_logging_func():
     def get_data(
         self, symbol: str, interval: str, start_datetime: pd.Timestamp, end_datetime: pd.Timestamp
     ) -> pd.DataFrame:
-        @deco_retry(retry_sleep=self.delay)
+        if hasattr(self, 'retry_config'):
```
you-n-g, line 151: Will it look simpler  to make retry as a part of the class  interface ( for example, [abstract  attribute](https://stackoverflow.com/questions/2736255/abstract-attributes-in-python) )

scripts/data_collector/utils.py
igor17400: Yes, much better!
It was a lack of attention on my part

scripts/data_collector/yahoo/collector.py
igor17400: @you-n-g but should I define this class interface only for `YahooCollectorBR1d` and `YahooCollectorBR1min` classes?

you-n-g: Maybe we can define

```
class YahooCollector(...):
      retry = 5  # Configuration attribute.  How many times will it try to re-request the data if the network fails.

class YahooCollectorBR(...):
        """"
            The reason to use retry=2 is due to the fact that
            Yahoo Finance unfortunately does not keep track of the majority
            of Brazilian stocks. 
            
            Therefore, the decorator deco_retry with retry argument
            set to 5 will keep trying to get the stock data 5 times, 
            which makes the code to download Brazilians stocks very slow. 
            
            In future, this may change, but for now 
            I suggest to leave retry argument to 1 or 2 in 
            order to improve download speed.
            In order to achieve this code logic an argument called retry_config
            was added into YahooCollectorBR base class
        """
      retry = 2
```

Now the retry is part of the interface, so we don't have to use hasattr to access the retry.

igor17400: @you-n-g I've added such modifications into commit https://github.com/microsoft/qlib/pull/990/commits/c31380481d189d5b7e05490dd9c662f975dffa97

you-n-g:
It looks great now!

SunsetWolf:
Hi, @igor17400 
This script is not working right now because of the missing datasource, hopefully the datasource can be updated.

Issue #988, opened 2022-03-17
How to query the data of the historical reporting period when using PIT data?


Chaoyingz:
## 🌟 Feature Description

How to query the data of the historical reporting period when using PIT data?
Is there a way to achieve this effect?
```python
>>> import qlib
>>> from qlib.data import D
>>> qlib.init()
>>> instruments = ["sh600519"]
>>> fields = ["P($$roewa_q)", "P($$yoyni_q)"]
>>> D.features(instruments, fields, start_time="2019-01-01", end_time="2020-01-01", freq="quarter")
                                  P($$roewa_q)  P($$yoyni_q)
        instrument datetime
        sh600519   2019-03-31      1.000000      1.305041
                   2019-06-30      2.000000      2.305041
                   2019-09-30      3.000000      3.305041
                   2019-12-31      4.175322      4.252650
```
If possible, use a period instead of datetime for indexing, like this:
```python
                       P($$roewa_q)  P($$yoyni_q)
instrument period
sh600519   201901      1.000000      1.305041
           201902      2.000000      2.305041
           201903      3.000000      3.305041
           201904      4.175322      4.252650
```
I am going to write a `PITPeriodProvider` to query the stored PIT data by referring to `ArcticFeatureProvider`. Is there a better way?
## Motivation

1. Application scenario:
2. Related works (Papers, Github repos etc.):
3. Any other relevant and important information:

you-n-g:
@Chaoyingz 
The value of PIT data depends on your _observation timestamp_, which is the meaning of `datetime` index in the current version. So I think we should consider 4 dimensions when we process PIT data  (i.e. <instrument,  datetime/observation time, period time, factor>).  Ignoring any of them will make the query vague.

If I want to refer to historical data , I'll use operators like  `"P(Ref($$roewa_q, 1))", P(Ref($$roewa_q, 2))", P(Ref($$roewa_q, 2))"`, and  its value will changes at different _observation timestamp_

Can you give us more details about the scenario you encountered?

Chaoyingz:
@you-n-g 
Thank you for your reply, My application scenario is to filter out the stock pool based on the historical financial data of the stock, so I need to use the current time as the observation point to query the data.

for example, I want to use the current time as the observation point to query the data with the reporting period of 201901. What should I do?

you-n-g:
> @you-n-g 
[...]

PR #1000, opened 2022-03-22
Add PRef operator (#988)

Chaoyingz:
## Description

Added `PRef` operator to support querying pit data for a specified reporting period.

## Motivation and Context

See #988 .

## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:
image

## Types of changes

- [ ] Fix bugs
- [x] Add new feature
- [ ] Update documentation


you-n-g:
Hi,  @Chaoyingz
This implementation looks really clean and smart!
But I think I misled you by giving [the example ](https://github.com/microsoft/qlib/issues/988#issuecomment-1074797066) 

I come up with another idea to make it implemented with less code (PRef may not be necessary. We just need to enhance the PFeature)
https://github.com/Chaoyingz/qlib/pull/2

Looking forward for your comments about it

Chaoyingz:
Hi, @you-n-g 
I think `PFeature` should be a feature rather than an expression, no additional parameters should be added to the feature.
How about extending an existing `P` expression directly? like `P($$roewa_q, 201901)`.

you-n-g:
@Chaoyingz  
I think both implementations are good enough but not perfect.
- The previous version will make `PFeature` more complex
- The current version will make `load` function more complex. 

We are trying to introduce minimal complexity (fewer methods, fewer arguments in the interface) into the framework when adding new features.
Instead, if the complexity is added to a concrete subclass, it would keep the framework clean.

And I come up with another option just now. How about adding a subclass for `PFeature` which support an extra `period` argument?  (e.g. changing PF to such a subclass from an alias)


I think all the above solutions are good enough. You can choose the one you like :)

Chaoyingz:
@you-n-g 
I think `PFeature` is a feature (corresponding to the file name), and `P` is an operator. If you want to build a feature dynamically, it is better to use an operator. So pref is the better choice.

you-n-g:
Hi， @Chaoyingz  
I agree with you
Please add more docs for `load` https://github.com/microsoft/qlib/blob/main/qlib/data/base.py#L157 to describe the `period` argument.

I think your next step looks very reasonable.
By such design, maybe the performance of PIT be boosted a lot.

Chaoyingz:
@you-n-g 👌

you-n-g:
It looks great!

PR #1049, opened 2022-04-14
Use the region in qlib.config for FileCalendarStorage.

ChiahungTai:
## Description

1. Add parameter region in method "resam_calendar".
2. Add region in time.py
3. Get region in qlib.config in FileCalendarStorage

## Motivation and Context

The high frequency backtest use the calendar for minutes data. Once the region is not "cn", the result of resam_calendar will be now due to differnt trasaction time in different region.

## How Has This Been Tested?
<! ---  Put an `x` in all the boxes that apply: --->

1. Pipeline test:
4. Your own tests:
Screen Shot 2022-04-14 at 9 14 45 PM

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/utils/resam.py
+++ b/qlib/utils/resam.py
@@ -8,8 +8,9 @@ from . import lazy_sort_index
 from .time import Freq, cal_sam_minute
 
 
-def resam_calendar(calendar_raw: np.ndarray, freq_raw: Union[str, Freq], freq_sam: Union[str, Freq],
-                   region: str) -> np.ndarray:
+def resam_calendar(
+    calendar_raw: np.ndarray, freq_raw: Union[str, Freq], freq_sam: Union[str, Freq], region: str
+) -> np.ndarray:
     """
     Resample the calendar with frequency freq_raw into the calendar with frequency freq_sam
     Assumption:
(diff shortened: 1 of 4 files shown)


ChiahungTai:
@you-n-g Could you please review this patch?

qlib/utils/resam.py
```diff
@@ -8,7 +8,9 @@
 from .time import Freq, cal_sam_minute
 
 
-def resam_calendar(calendar_raw: np.ndarray, freq_raw: Union[str, Freq], freq_sam: Union[str, Freq]) -> np.ndarray:
+def resam_calendar(
+    calendar_raw: np.ndarray, freq_raw: Union[str, Freq], freq_sam: Union[str, Freq], region: str
```
you-n-g, line 12: I think we can make the region an optional parameter.
And the default value is C['region'] if it is None

Otherwise, all the places need explicitly values when call this function.
image

qlib/utils/time.py
```diff
@@ -236,11 +271,16 @@ def time_to_day_index(time_obj: Union[str, datetime], region: str = "cn"):
             return int((time_obj - US_TIME[0]).total_seconds() / 60)
         else:
             raise ValueError(f"{time_obj} is not the opening time of the {region} stock market")
+    elif region == "tw":
```
you-n-g, line 274: REG_TW & REG_US to replace the string.

you-n-g:
It looks great!
Thanks so much!

Issue #1011, opened 2022-03-25
[Proposal] Systematic RL support in qlib


ultmaster:
2022/2/25

Package name:

* qlib.neutrader?
    * Sound, brand
    * Sounds like limited to "trading" scenario
* qlib.rl?
    * Shorter, easier to remember
    * Not exactly an RL framework
    * No ML as opposed to RL.
* qlib.sdm?
    * What is sdm?

K: Keep in an internal repo (possibly another repo)

D: Delete

TBD: To be discussed (possibly need merge effort)

Major efforts

* Unify config
    * Currently neutrader is based on `utilsd.config`.
    * qlib has another configuration system.
* Unify entry
    * Currently neutrader has several command line tools
    * qlib is mostly Pythonic launching.
    * Merge with workflow, qrun.
* Documentation and tests
    * Neutrader is almost zero-documented.
    * Neutrader has its own pytest.
    * … and coveragerc.

<table>
  <tr>
   <td>
K
   </td>
   <td>.azure/docker.yml
   </td>
  </tr>
  <tr>
   <td>D
   </td>
   <td>.azure/pipeline.yml
   </td>
  </tr>
  <tr>
   <td>TBD
   </td>
   <td>.coveragerc
   </td>
  </tr>
  <tr>
   <td>K
   </td>
   <td>.gitignore
   </td>
  </tr>
  <tr>
   <td>K
   </td>
   <td>README.md
   </td>
  </tr>
  <tr>
   <td>K
   </td>
   <td>docker/base.Dockerfile
   </td>
  </tr>
  <tr>
   <td>K
   </td>
   <td>docker/neutrader.Dockerfile
   </td>
  </tr>
  <tr>
   <td>qlib/examples
   </td>
   <td>examples/backtest/qlib.yml
   </td>
  </tr>
  <tr>
   <td>qlib.contrib.data.utils
   </td>
   <td>examples/data/ordergen.py
   </td>
  </tr>
  <tr>
   <td>
   </td>
   <td>neutrader/__init__.py
   </td>
  </tr>
  <tr>
   <td>&lt;neutrader>.action
   </td>
   <td>neutrader/action.py
   </td>
  </tr>
  <tr>
   <td>
   </td>
   <td>neutrader/data/__init__.py
   </td>
  </tr>
  <tr>
   <td>neutrader.data
   </td>
   <td>neutrader/data/base.py
   </td>
  </tr>
  <tr>
   <td>neutrader.utils
   </td>
[...]

PR #1076, opened 2022-04-25
Qlib RL framework (stage 1) - single-asset order execution

ultmaster:
This is the first stage of a systematic Qlib RL support (aka NeuTrader), as planned in #1011.

image

What this PR includes:

- Basic framework: simulator, state/action-interpreter, reward, logger.
- Utilities: data queue, finite env, env wrapper.
- The first single-asset order execution simulator built upon "OPD-styled" data, along with several interpreters and basic policies.
- Tests. Unit-tests and an end-to-end runnable tests for TWAP strategy.

What this PR will include but haven't been completed so far (we can discuss about whether they can be deferred to stage 2):

- [X] [MUST DO] Put the test data on the shared storage, and check the correctness of several policy checkpoints, against the old framework.
- [ ] The script to train a new policy. -- will be in stage 2

What this PR won't include:

- More loggers including tensorboard, mlflow, memory buffer.
- Fix the compatibility with non-linux platforms.
- Integration with Qlib workflow.
- Integration with Qlib config system - launching backtest / training via config.
- The "RLStrategy" that enables testing a trained policy in qlib.backtest for deployment.
- The second SAOE simulator that is built upon qlib.backtest.
- Performance optimization: prefetching data, caching.
- Multi-agent.
- Tasks other than SAOE.

As this PR is already quite large by itself, I suggest reviewing it first, and collecting some feedbacks.

**Please squash this PR when merging. The commit history is too messy.**


ultmaster:
image

tests/rl/test_saoe_simple.py
```diff
@@ -0,0 +1,306 @@
+# Copyright (c) Microsoft Corporation.
```
you-n-g, line 1: It raises error when running test with pandas-1.3.5

image

you-n-g: it raises less error when running test with pandas-1.4.2

image

you-n-g: numpy from 1.21.2 to  1.21.6 does not help
image

setup.py
```diff
@@ -47,39 +48,7 @@ def get_version(rel_path: str) -> str:
...
-    "joblib>=0.17.0",
-    "ruamel.yaml>=0.16.12",
-    "pymongo==3.7.2",  # For task management
-    "scikit-learn>=0.22",
-    "dill",
-    "dataclasses;python_version<'3.7'",
-    "filelock",
-    "jinja2<3.1.0"  # for passing the readthedocs workflow.
-]
+REQUIRED = [str(req) for req in parse_requirements(open("requirements.txt"))]
```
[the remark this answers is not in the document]
you-n-g: `read` returns the content and `open` return an IO.  
The suggestion needs further coding.

tests/rl/test_saoe_simple.py
you-n-g: install torch from 1.21.2  to 1.21.6 does not help
image

you-n-g: @ultmaster   Do you have any suggestions about how to successfully run the test successfully locally? 

image

requirements.txt
```diff
@@ -0,0 +1,32 @@
...
+lightgbm>=3.3.0
+tornado
+joblib>=0.17.0
+ruamel.yaml>=0.16.12
+pymongo==3.7.2  # For task management
+scikit-learn>=0.22
+dill
+dataclasses; python_version < "3.7"
+filelock
+jinja2<3.1.0  # for passing the readthedocs workflo
```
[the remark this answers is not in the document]
matluster: I think you are reviewing an outdated version of this PR?

tests/rl/test_saoe_simple.py
matluster: That's strange. What's you tianshou version?

Could you post the full content of `.output/result.csv`?

you-n-g: my `tianshou==0.4.8`:  
same as [the CI](https://github.com/microsoft/qlib/runs/6340717318?check_suite_focus=true#step:12:208)


Here is my output of  `pip freeze`
[env.txt](https://github.com/microsoft/qlib/files/8666117/env.txt)

Here is my `./tests/rl/.output/result.csv`
[result.csv](https://github.com/microsoft/qlib/files/8666145/result.csv)

@matluster

you-n-g:
@matluster  @ultmaster  Do we have a plan about the docs of the RL component? 
.... and,  are both of you the same person?

matluster:
Both are me. My computer gets locked out from the other account because of some Microsoft policy. I can only access the account on my cellphone...

About docs: no currently. I don't have plans to announce the feature officially, at least not before the training interface is ready.

tests/rl/test_saoe_simple.py
matluster: Is this the only failed test? Or some other tests also failed?

matluster: Could you print self.twap_price here, and see whether it equals to `market_price.mean()`?

image

If not, try to see what's `self.ticks_for_oder` here.

image

qlib/rl/order_execution/simulator_simple.py
```diff
@@ -0,0 +1,393 @@
...
+        self.position -= exec_vol.sum()
+        if self.position < -EPS or (exec_vol < -EPS).any():
+            raise ValueError(f"Execution volume is invalid: {exec_vol} (position = {self.position})")
+
+        # Get time index available for this step
+        time_index = self._get_ticks_slice(self.cur_time, self._next_time())
+
+        self.history_exec = self._dataframe_append(
+            self.history_exec,
+            SAOEMetrics(
```
lihuoran, line 211: Many parameters are unexpected.

qlib/rl/order_execution/policy.py
```diff
@@ -0,0 +1,158 @@
...
+# ppo #
+
+
+class PPOActor(nn.Module):
+    def __init__(self, extractor: nn.Module, action_dim: int):
+        super().__init__()
+        self.extractor = extractor
+        self.layer_out = nn.Sequential(nn.Linear(cast(int, extractor.output_dim), action_dim), nn.Softmax(dim=-1))
+
+    def forward(self, obs, state=None, info={}):
```
lihuoran, line 56: Default argument value is mutable. The better way is:
```python
def forward(self, obs, state=None, info=None):
    if info is None:
        info = {}
```

qlib/rl/utils/env_wrapper.py
```diff
@@ -0,0 +1,234 @@
...
+
+    cur_step: int
+    done: bool
+    initial_state: Any | None
+    obs_history: list
+    action_history: list
+    reward_history: list
+
+
+class EnvWrapper(
```
lihuoran, line 51: `EnvWrapper` is only used for wrapping `gym.Env`?

qlib/rl/utils/env_wrapper.py
```diff
@@ -0,0 +1,234 @@
...
+    gym.Env[ObsType, PolicyActType], Generic[InitialStateType, StateType, ActType, ObsType, PolicyActType]
```
lihuoran, line 52: Where should we put `Generic`? At the beginning or at the end?

qlib/rl/entries/test.py
```diff
@@ -0,0 +1,82 @@
...
+        Type of finite env implementation.
+    concurrency
+        Parallel workers.
+    """
+
+    # To save bandwidth
+    min_loglevel = min(lg.loglevel for lg in logger) if isinstance(logger, list) else logger.loglevel
+
+    with DataQueue(initial_states) as seed_iterator:
+        vector_env = finite_env_factory(
```
lihuoran, line 63: `finite_env_factory` takes `Callable[..., gym.Env]` as input but here provides a `Callable[[], EnvWrapper]`. What is the relationship between `gym.Env` with `EnvWrapper`?

qlib/rl/data/pickle_styled.py
```diff
@@ -0,0 +1,251 @@
...
+
+
+class IntradayProcessedData:
+    """Processed market data after data cleanup and feature engineering.
+
+    It contains both processed data for "today" and "yesterday", as some algorithms
+    might use the market information of the previous day to assist decision making.
+    """
+
+    today: pd.DataFrame
```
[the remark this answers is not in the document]
matluster: It's mainly for typehints and docstrings.

qlib/rl/utils/data_queue.py
```diff
@@ -0,0 +1,166 @@
...
+        if queue_maxsize == 0:
+            if os.cpu_count() is not None:
+                queue_maxsize = cast(int, os.cpu_count())
+                _logger.info(f"Automatically set data queue maxsize to {queue_maxsize} to avoid overwhelming.")
+            else:
+                queue_maxsize = 1
+                _logger.warning(f"CPU count not available. Setting queue maxsize to 1.")
+
+        self.dataset: Sequence[T] = dataset
+        self.repeat: int = repeat
```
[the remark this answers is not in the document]
matluster: I think it's useful to expose them as public attributes.

qlib/rl/utils/data_queue.py
```diff
@@ -0,0 +1,166 @@
...
+        self._done = multiprocessing.Value("i", 0)
+
+    def __enter__(self):
+        self.activate()
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.cleanup()
+
+    def cleanup(self):
```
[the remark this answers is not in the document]
matluster: In that case, we can enable some settings like `reportMissingTypes` in linter, which will report all missing typehints.

qlib/rl/utils/env_wrapper.py
```diff
@@ -0,0 +1,234 @@
...
+    """
+    This is the status data structure used in EnvWrapper.
+    The fields here are in the semantics of RL.
+    For example, ``obs`` means the observation fed into policy.
+    ``action`` means the raw action returned by policy.
+    """
+
+    cur_step: int
+    done: bool
+    initial_state: Any | None
```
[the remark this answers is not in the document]
matluster: See [PEP 563](https://peps.python.org/pep-0563/) for the discussion.

qlib/rl/utils/log.py
```diff
@@ -0,0 +1,398 @@
...
+        self.log_step(rew, values)
+
+        if done:
+            # Update counter
+            self.global_episode += 1
+            self.episode_count += 1
+
+            self.log_episode(self.episode_lengths[env_id], self.episode_rewards[env_id], self.episode_logs[env_id])
+
+    def on_env_reset(self, env_id: int, obs: ObsType) -> None:
```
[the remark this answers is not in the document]
matluster: It's an interface for subclasses.

qlib/rl/aux_info.py
```diff
@@ -0,0 +1,43 @@
...
+
+__all__ = ["AuxiliaryInfoCollector"]
+
+AuxInfoType = TypeVar("AuxInfoType")
+
+
+class AuxiliaryInfoCollector(Generic[StateType, AuxInfoType]):
+    """Override this class to collect customized auxiliary information from environment."""
+
+    env: "EnvWrapper" | None = None
```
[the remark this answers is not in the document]
matluster: No we can't. Otherwise we will have circular imports.

qlib/rl/utils/env_wrapper.py
matluster: As suggested in docstring:
Qlib-based RL environment, subclassing ``gym.Env``. A wrapper of components, including simulator, state-interpreter, action-interpreter, reward.

qlib/rl/utils/env_wrapper.py
matluster: IDK. Any suggestions?

qlib/rl/entries/test.py
matluster: The `Callable[..., gym.Env]` type is passed to tianshou. I don't try to handle those.

gym.Env is an abstract env. EnvWrapper is a concrete implementation of `gym.Env` is to wrap our components (including simulator, interpreter, ...) into a `gym.Env` so that it can work in tianshou.

matluster:
**A little more background**.

The whole point of idea is: we want an existing RL framework (i.e., tianshou built upon gym), to provide backend support (including RPC capability, policy algos, training recipe) for our designs (including simulator, interpreter, reward...), so that we don't have to implement the whole stack of yet another RL framework.

So a natural questions is then: why do we put so much effort into these designs? Why can't we simply "use" the tianshou framework? The answer is in a few folds:

- In our scenario, trajectories run on data (it's not a random game). Thus we need FiniteEnv and DataQueue to ensure running through all the data exactly once (at least in backtest).
- See the loop in PR description (policy - action interpreter - simulator - state interpreter - policy). The interpreters are actually part of the *strategy* (see qlib strategy) to be deployed as an online service, but they should be wrapped into `gym.Env` if the policy is to be trained in a traditional RL framework.
- Explicitly breaking things apart makes them modular and easier to combine one with another. The prototype version of this design, as an internal project, has been proven to facilitate research over the past years.

qlib/rl/utils/data_queue.py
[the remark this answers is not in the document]
lihuoran: For my personal habit I would keep them private until they are truly needed from outside. But it is okay to leave it as what it is now~

qlib/rl/utils/data_queue.py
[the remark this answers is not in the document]
lihuoran: Agree with it.

qlib/rl/aux_info.py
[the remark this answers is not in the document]
lihuoran: I think it will not trigger circular imports because `from .utils.env_wrapper import EnvWrapper` is wrapped by `if TYPE_CHECKING:`. The import operation will not be executed in runtime.

qlib/rl/utils/env_wrapper.py
lihuoran: From [this doc](https://mypy.readthedocs.io/en/stable/generics.html#defining-sub-classes-of-generic-classes) maybe we can put it at the end.

lihuoran:
@matluster I use PyCharm as my local IDE and PyCharm uses PEP8 to inspect code. Some of the issues will be reported while editing rather than being reported by mypy. The other issues are majorly my personal experiences, so it is fine if the team think they are unnecessary.

qlib/rl/order_execution/simulator_simple.py
lihuoran: Sorry I didn't make myself clear. What I was trying to mean is there are some type mismatch issues. For example, the first argument is `stock_id=self.order.stock_id`, however, `SAOEMetrics.stock_id` is a `pd.Timestamp` but `self.order.stock_id` is a str.

qlib/rl/order_execution/simulator_simple.py
```diff
@@ -0,0 +1,393 @@
...
+    """This price is used to compute price advantage.
+    It"s defined as the average price in the period from order"s start time to end time."""
+
+    ticks_index: pd.DatetimeIndex
+    """All available ticks for the day (not restricted to order)."""
+
+    ticks_for_order: pd.DatetimeIndex
+    """Ticks that is available for trading (sliced by order)."""
+
+    def __init__(
```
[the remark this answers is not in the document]
lihuoran: We can discuss this issue later: shall we call `__init__` of super class even if logically we do not need to do so.

qlib/rl/utils/data_queue.py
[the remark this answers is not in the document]
matluster: Report missing types will generate this many errors:

image

Most of them are missing annotations of `self`. I'd rather not enable this...

qlib/rl/order_execution/simulator_simple.py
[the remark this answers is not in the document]
matluster: They should. But they shouldn't when the super class is actually an interface.

Yes, `Simulator` is an interface. It didn't subclass ABC because I try to avoid abusing ABC (see [PEP-3119](https://peps.python.org/pep-3119/)).

qlib/rl/order_execution/simulator_simple.py
matluster: stock_id should be a str. I don't know why mypy didn't complain about this.

matluster:
> @matluster I use PyCharm as my local IDE and PyCharm uses PEP8 to inspect code. Some of the issues will be reported while editing rather than being reported by mypy. The other issues are majorly my personal experiences, so it is fine if the team think they are unnecessary.

I failed to see how pep8 can generate those messages you listed above.

https://pep8.readthedocs.io/en/release-1.7.x/intro.html#error-codes

qlib/rl/utils/data_queue.py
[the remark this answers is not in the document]
lihuoran: Shall we just report `missing a return type annotation`?

qlib/rl/order_execution/simulator_simple.py
lihuoran: Yes, and many parameters have unexpected types. Mypy doesn't complain but my IDE does XD.

qlib/rl/order_execution/simulator_simple.py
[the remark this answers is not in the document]
lihuoran: Are there any disadvantages to call `__init__` when its super class is an interface? In my understanding, a class is an "interface" means it has only methods but no attributes, so I didn't get the point about why we shouldn't call its `__init__`.

For the ABC part, is it OK to use `metaclass=abc.ABCMeta`? I am not sure about this part, so it is up to you~

lihuoran:
> > @matluster I use PyCharm as my local IDE and PyCharm uses PEP8 to inspect code. Some of the issues will be reported while editing rather than being reported by mypy. The other issues are majorly my personal experiences, so it is fine if the team think they are unnecessary.
> 
> I failed to see how pep8 can generate those messages you listed above.
> 
> https://pep8.readthedocs.io/en/release-1.7.x/intro.html#error-codes

Not all of the issues I mentioned above are reported by PEP8. Some of them (for example, the mismatched parameter type issue) are reported by the internal checker of PyCharm. I think for the code format issues, we should NOT reply on the type of IDEs, so we could see if we can find an elegant way to enrich the current format checkers at a proper level. For "IDE related" issues, I will report them case by case and we could discuss whether they should be fixed.

It may take me a while to get familiar with our coding standards. After that, I should have a better grasp of which issues should be addressed and which ones should be ignored.

tests/rl/test_saoe_simple.py
you-n-g: @ultmaster @matluster 
I'm trying to debug it.
First, I try to change it into a single process instead of multi-processing.
So, I made the following changes for DummyVectorEnv

image


But I got the following error when I ran it again.

Is it the correct way to debug the program? 

image

qlib/rl/order_execution/simulator_simple.py
matluster: Just added a comment.

qlib/rl/order_execution/simulator_simple.py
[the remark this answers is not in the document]
matluster: The interface could have `NotImplementedError()`. We definitely shouldn't call super() when it contains `NotImplementedError()`.

I don't like `abc.ABCMeta`. It breaks sometimes when we need to customize metaclass.

qlib/rl/data/pickle_styled.py
```diff
@@ -0,0 +1,251 @@
+# Copyright (c) Microsoft Corporation.
```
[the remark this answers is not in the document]
matluster: I think qlib.rl is a self-contained package, and pickle-styled data is the only type of supported data format. Let's keep it here for now.

qlib/rl/utils/data_queue.py
```diff
@@ -0,0 +1,166 @@
...
+            self._done.value += 1
+        for repeat in range(500):
+            if repeat >= 1:
+                warnings.warn(f"After {repeat} cleanup, the queue is still not empty.", category=RuntimeWarning)
+            while not self._queue.empty():
+                try:
+                    self._queue.get(block=False)
+                except Empty:
+                    pass
+            time.sleep(1.0)  # a chance to allow more data to be put in
```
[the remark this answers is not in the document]
matluster: Explained in comments.

tests/rl/test_saoe_simple.py
matluster: Nice catch. It was a flaw in this design. I added a hot-fix so that dummy env could work.

Now the tests have passed on my side. Please try again.

matluster:
> > > @matluster I use PyCharm as my local IDE and PyCharm uses PEP8 to inspect code. Some of the issues will be reported while editing rather than being reported by mypy. The other issues are majorly my personal experiences, so it is fine if the team think they are unnecessary.
> > 
> > 
> > I failed to see how pep8 can generate those messages you listed above.
> > https://pep8.readthedocs.io/en/release-1.7.x/intro.html#error-codes
> 
> Not all of the issues I mentioned above are reported by PEP8. Some of them (for example, the mismatched parameter type issue) are reported by the internal checker of PyCharm. I think for the code format issues, we should NOT reply on the type of IDEs, so we could see if we can find an elegant way to enrich the current format checkers at a proper level. For "IDE related" issues, I will report them case by case and we could discuss whether they should be fixed.
> 
> It may take me a while to get familiar with our coding standards. After that, I should have a better grasp of which issues should be addressed and which ones should be ignored.

I think it could be another task: adding another linter to qlib CI.

Let's keep as is for now. This PR is already super fat. :)

lihuoran:
> > > > @matluster I use PyCharm as my local IDE and PyCharm uses PEP8 to inspect code. Some of the issues will be reported while editing rather than being reported by mypy. The other issues are majorly my personal experiences, so it is fine if the team think they are unnecessary.
> > > 
> > > 
> > > I failed to see how pep8 can generate those messages you listed above.
> > > https://pep8.readthedocs.io/en/release-1.7.x/intro.html#error-codes
> > 
> > 
> > Not all of the issues I mentioned above are reported by PEP8. Some of them (for example, the mismatched parameter type issue) are reported by the internal checker of PyCharm. I think for the code format issues, we should NOT reply on the type of IDEs, so we could see if we can find an elegant way to enrich the current format checkers at a proper level. For "IDE related" issues, I will report them case by case and we could discuss whether they should be fixed.
> > It may take me a while to get familiar with our coding standards. After that, I should have a better grasp of which issues should be addressed and which ones should be ignored.
> 
> I think it could be another task: adding another linter to qlib CI.
> 
> Let's keep as is for now. This PR is already super fat. :)

Sure~
Please pay attention to the logic parts and we can leave the format parts to future work.

you-n-g:
@ultmaster  @matluster  https://github.com/ultmaster/qlib/pull/1 Could you please merge this PR?

qlib/rl/utils/finite_env.py
```diff
@@ -0,0 +1,315 @@
...
+            self._collector_guarded = False
+
+        # At last trigger the loggers
+        for logger in self._logger:
+            logger.on_env_all_done()
+
+    def reset(self, id=None):
+        assert not self._zombie
+
+        if not self._collector_guarded:
```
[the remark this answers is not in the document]
you-n-g: I didn't find any [usage of this variable in tianshou](https://github.com/thu-ml/tianshou/search?q=_collector_guarded). 

And when I grep this variable.  
It does not affect the running workflow of the program  (only one `if` statement is affected).  It just maintains this variable.
image

Will it be used in the future?

tests/rl/test_saoe_simple.py
you-n-g: @ultmaster  @matluster  
Thanks for your hot-fix

I try to debug.  Here is some information I got.

1)  I start the test by `pytest -s --pdb --disable-warnings /sdc/home/xiaoyang/re
  pos/qlib-main/libs/qlib/tests/rl/test_saoe_simple.py::test_twap_strategy`
It create 4 `DummyVectorEnv` and here is the first one.
It uses the deal price **all day** to get  twap_price.
image

2) I skip the next 3 initialization of `DummyVectorEnv` and stop at the first step in the first `DummyVectorEnv`
It uses the **first 30 tick market price** to get PA in the first step
image

IIUC,  the PA at this step will not be zero (though the average PA all day will be 0).

But It tries to assert that the PA of **every step** is zero.  I don't think it is reasonable.
image

qlib/rl/entries/test.py
```diff
@@ -59,16 +60,32 @@ def backtest(
     # To save bandwidth
     min_loglevel = min(lg.loglevel for lg in logger) if isinstance(logger, list) else logger.loglevel
 
+    def env_factory():
+        # FIXME: state_interpreter and action_interpreter are stateful (having a weakref of env),
+        # and could be thread unsafe.
+        # I'm not sure whether it's a design flaw.
+        # I'll rethink about this when designing the trainer.
+
+        if finite_env_type == "dummy":
```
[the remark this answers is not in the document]
matluster: I always prefer Literal.

The enum is only used when the value can't speak for itself and shouldn't be used directly. For example, log-level equals 10. You don't really know what 10 means, and 10 is not easy to remember either.

qlib/rl/utils/env_wrapper.py
```diff
@@ -0,0 +1,237 @@
...
+        State-observation converter.
+    action_interpreter
+        Policy-simulator action converter.
+    seed_iterator
+        An iterable of seed. With the help of :class:`qlib.rl.utils.DataQueue`,
+        environment workers in different processes can share one ``seed_iterator``.
+    reward_fn
+        A callable that accepts the StateType and returns a float (at least in single-agent case).
+    aux_info_collector
+        Collect auxiliary informations. Could be useful in MARL.
```
you-n-g, line 79:
```suggestion
        Collect auxiliary information. Could be useful in MARL.
```

qlib/rl/utils/finite_env.py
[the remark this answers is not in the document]
matluster: This variable is used to avoid misuse: calling collect without collector_guard. What collector guard really does is in L189-L198.

tests/rl/test_saoe_simple.py
matluster: We didn't assert every PA at every **step** to be zero. Rather, we assert PA for every **episode** to be zero. There is a difference.

What you are referring to at L228 is the PA at every step. We actually never checked that (it's not even logged as a metric actually). What I'm interested in is the PA at L351.

I checked you result.csv, `market_price` and `trade_price` (at L351), which PA relies on, are both correct (and equal, because it's a TWAP strategy), but PA is wrong. One interesting fact is that the `market_price` here is actually "re-calculated", and not `self.twap_price`. I suggest something odd could happen here.

So my suggestion is to add the following lines at L351:

```python
print(self.twap_price, market_price.mean(), np.sum(market_price * exec_vol))
```

You would see three **equal** values on each line. If you didn't, something is wrong.

(But on the other hand, I see your `twap_price` is calculated on the whole-day price, which is actually... correct in this case. So I'm confused here.)

you-n-g: @matluster @ultmaster  [This solution ](https://github.com/ultmaster/qlib/pull/2) will solve the problem in my system.
Please merge it

qlib/rl/utils/env_wrapper.py
```diff
@@ -0,0 +1,237 @@
...
+            if obj is not None:
+                obj.env = weakref.proxy(self)  # type: ignore
+
+        self.simulator_fn = simulator_fn
+        self.state_interpreter = state_interpreter
+        self.action_interpreter = action_interpreter
+
+        if seed_iterator is None:
+            # in this case, there won't be any seed for simulator
+            self.seed_iterator = SEED_INTERATOR_MISSING
```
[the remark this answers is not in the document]
matluster: Comments added.

qlib/rl/utils/env_wrapper.py
```diff
@@ -0,0 +1,237 @@
...
+        simulator_fn: Callable[..., Simulator[InitialStateType, StateType, ActType]],
+        state_interpreter: StateInterpreter[StateType, ObsType],
+        action_interpreter: ActionInterpreter[StateType, PolicyActType, ActType],
+        seed_iterator: Iterable[InitialStateType] | None,
+        reward_fn: Reward | None = None,
+        aux_info_collector: AuxiliaryInfoCollector[StateType, Any] | None = None,
+        logger: LogCollector | None = None,
+    ):
+        # assign weak reference to wrapper
+        for obj in [state_interpreter, action_interpreter, reward_fn, aux_info_collector]:
```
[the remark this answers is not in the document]
matluster: Comments added.

Saihan284 (approved):

PR #1118, opened 2022-06-02
fix off by one bug: when end day is the last date in calendar

NotF404:
when we backtest, if the end of backtest date is the last date in calendar, [calendar_index + 1] willbe out of index in calendar.

## Description

since just one line changed, just as the code  there is.

## Motivation and Context

it's a bug

## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/backtest/utils.py
+++ b/qlib/backtest/utils.py
@@ -114,7 +114,7 @@ class TradeCalendarManager:
             trade_step = self.get_trade_step()
         trade_step = trade_step - shift
         calendar_index = self.start_index + trade_step
-        return self._calendar[calendar_index], epsilon_change(self._calendar[calendar_index + 1])
+        return self._calendar[calendar_index], epsilon_change(self._calendar[calendar_index] + pd.Timedelta(days=1))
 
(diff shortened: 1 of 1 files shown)


lzh222333:
In this method, we do not think it is a bug. Because this method will not only used in daily trading, but also used in high-frequency trading. For your question, could you please give me more information about your task configuration? We think this class use "_calendar = Cal.calendar(freq=freq, future=True)", so it will use day_future.txt instead of day.txt. So please make sure your day_future.txt has enough reserved dates after your last predicted date. When your day_future.txt is right, the [calendar_index + 1] will not be out of index.

SunsetWolf:
The pylint error in CI has been fixed, try using `git pull` to resolve the pylint error in CI.

you-n-g:
@NotF404  Hi,  Thanks for your PR.  Did @lzh222333 's reply solve your question?

NotF404:
In my case,

Issue #1249, opened 2022-08-10
Support for storing str type data


Chaoyingz:
## 🌟 Feature Description

Support for storing str type data.

## Motivation

1. Application scenario
2. Related works (Papers, Github repos etc.):
3. Any other relevant and important information:

Store data such as name, industry, and block of stocks.

## Alternatives
This data is a bit like the `category` type in the `pandas` library, we can store the string in a list and then store its index.

Issue #232, opened 2021-01-31
Categorical features and cross-sectional neutralization


yieldbook:
1. The feature bin files only support float data now, but sometimes we need to dump categorical or string features(like industry or sector) into the feature bin files. 
2. Cross-sectional neutralization is a widely used data processor in equity quant investment. Maybe you should consider adding a neutralization processor, with which we can neutralize the data on sector, size or beta.

PR #1254, opened 2022-08-17
Add support for category data type

Chaoyingz:
## Description

See https://github.com/microsoft/qlib/issues/1249.
Supports storing data of string or category type.

If a column is of type string or category, then the value of the column will be stored as follows step:
1. Stores a list of unique values for this column(path is `qlib_dir/categories/column_name.txt`).
2. Convert the value of this column to the index of the value in the previous step list.
3. The column is stored in bin format.

You can query the value of the string or category type column with the following methods:
1. Query Index: `D.features(instruments, ["$column"])`
2. Query Value(use **Cat** operator): `D.features(instruments, ["Cat($column")])`

For specific usage, please refer to the test case(`tests/test_category_data.py`).

## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:
image

## Types of changes

- [ ] Fix bugs
- [x] Add new feature
- [ ] Update documentation

## Notes
related ISSUE https://github.com/microsoft/qlib/issues/232.

### DIFF
--- a/qlib/data/ops.py
+++ b/qlib/data/ops.py
@@ -5,6 +5,8 @@
 from __future__ import division
 from __future__ import print_function
 
+import pathlib
+
 import numpy as np
 import pandas as pd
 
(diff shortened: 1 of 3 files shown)


qlib/data/ops.py
```diff
@@ -118,6 +151,8 @@ def __init__(self, feature, func):
 
     def _load_internal(self, instrument, start_index, end_index, *args):
         series = self.feature.load(instrument, start_index, end_index, *args)
+        if series.dtype != np.float32:
```
you-n-g, line 154: This checker may be required in PairOperator,  Rolling, and PairRolling.
The checking can happen during the stage of initialization if operations like 'cat' are kept.

scripts/dump_bin.py
```diff
@@ -45,6 +47,7 @@ def __init__(
         exclude_fields: str = "",
         include_fields: str = "",
         limit_nums: int = None,
+        is_convert_category: bool = False,
```
you-n-g, line 50: Why do we have to use this parameter to control whether to convert the string to index? 

Will it be simpler if we only convert it before dumping it to `.bin` files?

qlib/data/ops.py
```diff
@@ -96,6 +98,37 @@ def _load_internal(self, instrument, start_index, end_index, *args):
         return self.feature.load(instrument, start_index, end_index, *args)
 
 
+class Cat(ElemOperator):
```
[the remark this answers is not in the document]
Chaoyingz: > Did you encounter any performance issues if we force convert it to category every time?

No performance issue. The reason for adding the `Cat` operator is that every time you take any field, you need to determine whether the field is of category type, which affects non-category type fields. 

It is a good suggestion to handle category data logic in `FileFeatureStorage`, which will make filestorage clearer.

Adding the function to get the field type in `FeatureProvider` does not seem to have practical use at present, and it will be added when it is useful.

qlib/data/ops.py
```diff
@@ -34,6 +34,9 @@
 
 
 #################### Element-Wise Operator ####################
+FLOAT_DTYPES = (float, np.float, np.float32, np.float64)
+CATEGORY_DTYPES = (str, "category")
+
```
you-n-g, line 39:
```suggestion


```

PR #1273, opened 2022-08-31
fix dump_bin

lerit:
dump_fix data that not in calendar_list, throw error:
```
NaT is not in list
```

if first data with **dump_all** like:
```
date,open
2000-01-04,111
2000-01-05,222
2000-01-06,333
```
and then second with **dump_fix** like:
```
date,open
2000-01-01,111
2000-01-02,222
2000-01-03,333
```
it will cause **_df** to be empty,and cause error(because df not in calendar_list range):
```
_df = self.data_merge_calendar(df, calendar_list)
```

**i know even this pull request ,old data still not be update to bin,it just a way to catch and give message for such data**

### DIFF
--- a/scripts/dump_bin.py
+++ b/scripts/dump_bin.py
@@ -219,6 +219,9 @@ class DumpDataBase:
             return
         # align index
         _df = self.data_merge_calendar(df, calendar_list)
+        if _df.empty:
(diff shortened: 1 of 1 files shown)


you-n-g:
Thanks!  It looks great.
Welcome to become one of [Qlib's contributors](https://github.com/microsoft/qlib/graphs/contributors)

PR #1275, opened 2022-09-01
refactor:invalid code

lerit:
as if the code is not in use

### DIFF
--- a/scripts/dump_bin.py
+++ b/scripts/dump_bin.py
@@ -476,9 +476,6 @@ class DumpDataUpdate(DumpDataBase):
                         futures[executor.submit(self._dump_bin, _df, _update_calendars)] = _code
                 else:
                     # new stock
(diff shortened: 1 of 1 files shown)


scripts/dump_bin.py
```diff
@@ -476,9 +476,6 @@ def _dump_features(self):
                         futures[executor.submit(self._dump_bin, _df, _update_calendars)] = _code
                 else:
                     # new stock
-                    _dt_range = self._update_instruments.setdefault(_code, dict())
```
you-n-g, line 479: This code will  act like
```
_dt_range = self._update_instruments[_code]
# and then do something on _dt_range 
```

Are you sure that it does nothing?

you-n-g: @lerit

lerit: sorry,my fault.

PR #1302, opened 2022-09-26
General handler for open source data preprocessing

Arthur-Null:
## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [ ] Fix bugs
- [x] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/contrib/data/highfreq_handler.py
+++ b/qlib/contrib/data/highfreq_handler.py
@@ -124,6 +124,7 @@ class HighFreqGeneralHandler(DataHandlerLP):
         day_length=240,
     ):
         self.day_length = day_length
+
         def check_transform_proc(proc_l):
             new_l = []
             for p in proc_l:
--- a/qlib/contrib/data/highfreq_handler.py
+++ b/qlib/contrib/data/highfreq_handler.py
@@ -110,7 +110,7 @@ class HighFreqHandler(DataHandlerLP):
         return fields, names
 
 
-class HighFreqOpenHandler(HighFreqHandler):
+class HighFreqGeneralHandler(HighFreqHandler):
     def __init__(
         self,
         instruments="csi300",
@@ -248,7 +248,7 @@ class HighFreqBacktestHandler(DataHandler):
 
         return fields, names
 
-class HighFreqOpenBacktestHandler(HighFreqBacktestHandler):
+class HighFreqGeneralBacktestHandler(HighFreqBacktestHandler):
     def __init__(
         self,
         instruments="csi300",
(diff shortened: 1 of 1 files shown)


qlib/contrib/data/highfreq_handler.py
```diff
@@ -163,6 +249,45 @@ def get_feature_config(self):
...
+            start_time=start_time,
+            end_time=end_time,
+        )
+
+    def get_feature_config(self):
+        fields = []
+        names = []
+
+        template_paused = f"Cut({{0}}, {self.day_length * 2}, None)"
+        # template_paused = "{0}"
```
you-n-g, line 272: What is this line for ?

qlib/contrib/data/highfreq_handler.py
```diff
@@ -110,6 +110,92 @@ def get_normalized_price_feature(price_field, shift=0):
         return fields, names
 
 
+class HighFreqGeneralHandler(HighFreqHandler):
```
you-n-g, line 113: Should this class inherit from  DataHandlerLP directly?

Both `class HighFreqHandler` and `class HighFreqGeneralHandler` are special cases of the high frequent handlers.
We don't have a base class for high frequent handlers. So it should inherit from DataHandlerLP directly

qlib/contrib/data/highfreq_handler.py
```diff
@@ -163,6 +249,45 @@ def get_feature_config(self):
         return fields, names
 
 
+class HighFreqGeneralBacktestHandler(HighFreqBacktestHandler):
```
you-n-g, line 252: Should this class inherit from DataHandlerLP directly?

Both class HighFreqHandler and class HighFreqGeneralHandler are special cases of the high frequent handlers.
We don't have a base class for high frequent handlers. So it should inherit from DataHandlerLP directly

qlib/contrib/data/highfreq_handler.py
```diff
@@ -124,14 +124,37 @@ def __init__(
         day_length=240,
     ):
         self.day_length = day_length
+
+        def check_transform_proc(proc_l):
```
you-n-g, line 128: This function duplicates with the previous one in this file. Pleases shared them

you-n-g, line 128: Why can't this function be used here 
https://github.com/microsoft/qlib/blob/a53f59cdf7f36d28f3998f2bb103c56d08cc2e63/qlib/contrib/data/handler.py#L11

PR #1321, opened 2022-10-19
optimize memory consumption

TruthZZ:
## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [ ] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/data/dataset/__init__.py
+++ b/qlib/data/dataset/__init__.py
@@ -323,11 +323,21 @@ class TSDataSampler:
         assert get_level_index(data, "datetime") == 0
         self.data = lazy_sort_index(data)
 
+        # the data type will be changed
+        # The index of usable data is between start_idx and end_idx
+        # move these forward so we can delete self.data earlier
+        self.idx_df, self.idx_map = self.build_index(self.data)
+        self.data_index = deepcopy(self.data.index)
+
         kwargs = {"object": self.data}
         if dtype is not None:
             kwargs["dtype"] = dtype
+            
+        self.data_arr = self.data.to_numpy(copy=True)   # Get index from numpy.array will much faster than DataFrame.values!
 
-        self.data_arr = np.array(**kwargs)  # Get index from numpy.array will much faster than DataFrame.values!
+        # del self.data  # save memory
+        self.data = self.data.drop(columns=self.data.columns, inplace=True) # del self.data won't work for fragile dataframe
+
         # NOTE:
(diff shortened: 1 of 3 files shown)


you-n-g:
@TruthZZ Please format your code with black to pass the CI.
https://github.com/microsoft/qlib/actions/runs/3281259496/jobs/5415910765#step:9:1

Thanks

qlib/data/dataset/handler.py
```diff
@@ -243,8 +243,9 @@ def _fetch_data(
                 data_df = fetch_df_by_col(data_df, col_set)
             else:
                 # Fetch column  first will be more friendly to SepDataFrame
-                data_df = fetch_df_by_col(data_df, col_set)
+                # Fetch index first will be more friendly to Memory. Because pandas will allocate new memory for the intermediate dataframe, and the extra dataframe will be smaller.
```
you-n-g, line 246: We have a  class like  this
https://github.com/microsoft/qlib/blob/main/qlib/contrib/data/utils/sepdf.py#L17

For this class, it is more friendly to fetch the column first.

So it is hard to decide which is better.
I suggest the following solutions (any one of them looks reasonable to me)
-  add a special method like the `default_fetch_order` method for  SepDataFrame and check 
-  Add `default_fetch_order` parameter for the handler and keep the default behavior the same as before.

qlib/data/dataset/__init__.py
```diff
@@ -358,7 +363,6 @@ def __init__(
         )
         self.idx_arr = np.array(self.idx_df.values, dtype=np.float64)  # for better performance
 
-        del self.data  # save memory
```
you-n-g, line 361: I think it will be better to delete this attribute even after you call self.data.drop.
Less attributes will make the interface simpler.

qlib/data/dataset/__init__.py
```diff
@@ -323,11 +323,21 @@ def __init__(
...
 
+        # the data type will be changed
+        # The index of usable data is between start_idx and end_idx
+        # move these forward so we can delete self.data earlier
+        self.idx_df, self.idx_map = self.build_index(self.data)
+        self.data_index = deepcopy(self.data.index)
+
         kwargs = {"object": self.data}
         if dtype is not None:
             kwargs["dtype"] = dtype
```
you-n-g, line 334: make sure `dtype` behaves same as before

PR #1329, opened 2022-10-26
Update cache.py

wony-zheng:
query with Feature object use disk_cache may cause error，example
  D.feature([symbol], [Feature('close')], disk_cache=1)


you-n-g:
It looks great! Thanks

PR #1332, opened 2022-10-27
Refine RL todos

lihuoran:
## Description

- Refine the return value of `collect_data_loop`: make it more structural & add more annotations. Refine the report generation logic in RL backtest accordingly.
- Decouple `Interpreter` with `Env`. Now, each interpreter has its own tick counter, so it does not need to get this information from `Env`. Remove `CollectDataEnvWrapper` since it is no longer needed.
- Move `SAOEStateAdapter` and related functions from `qlib/rl/order_execution/state.py` to `qlib/rl/order_execution/strategy.py`. (This was supposed to be done in [PR 1316](https://github.com/microsoft/qlib/pull/1316) but we somehow missed it).
- Change related test files.

## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [ ] Fix bugs
- [ ] Add new feature
- [ ] Update documentation


qlib/rl/utils/env_wrapper.py
```diff
@@ -129,9 +114,7 @@ def __init__(
         # 3. Avoid circular reference.
         # 4. When the components get serialized, we can throw away the env without any burden.
         #    (though this part is not implemented yet)
-        super().__init__()
-
-        for obj in [state_interpreter, action_interpreter, reward_fn, aux_info_collector]:
+        for obj in [reward_fn, aux_info_collector]:
```
ultmaster, line 117: In the original design, (1) state interpreter, (2) action interpreter, (3) reward function, (4) auxiliary info collector, are born equal and treated equally in the EnvWrapper. This change breaks such philosophy. Extra discussion might be needed for this.

qlib/rl/order_execution/strategy.py
```diff
@@ -221,20 +490,11 @@ def __init__(
...
-        # We should decompose the interpreters with EnvWrapper in the future and we should remove this method
-        # after that.
-
-        self._env = env
-        self._state_interpreter.env = self._action_interpreter.env = self._env
-
     def reset(self, outer_trade_decision: BaseTradeDecision = None, **kwargs: Any) -> None:
         super().reset(outer_trade_decision=outer_trade_decision, **kwargs)
 
-        # In backtest, env.reset() needs to be manually called since there is no outer trainer to call it
```
ultmaster, line 235: Why can't we simply create a `DummyEnvWrapper` here?

qlib/rl/utils/env_wrapper.py
ultmaster: Another reason that motivates the previous design is that I've seen some algorithms that need extra communication between components. For example, reward function might rely on extra states calculated by state interpreter to fasten its calculation. Giving them access to env wrapper makes such "hacking" possible.

qlib/rl/order_execution/strategy.py
```diff
@@ -15,22 +15,294 @@
...
+        start += granularity
+
+    if ret[-1] > end:
+        ret.pop()
+    if ret[-1] == end and not include_end:
+        ret.pop()
+    return pd.DatetimeIndex(ret)
+
+
+def fill_missing_data(
```
you-n-g, line 50: Hope https://github.com/microsoft/qlib/pull/1333 can help to remove this function.

qlib/rl/utils/env_wrapper.py
lihuoran: Understood. I reverted this change to let interpreters have `env` again. It's just that `env` won't be used at the moment.

qlib/rl/order_execution/strategy.py
lihuoran: In my personal understanding, we don't need a `DummyEnvWrapper` at least for now.

qlib/rl/order_execution/policy.py
```diff
@@ -148,7 +148,11 @@ def __init__(
             action_space=action_space,
         )
         if weight_file is not None:
-            load_weight(self, weight_file)
+            loaded_weight = torch.load(weight_file, map_location="cpu")
+            if "vessel" in loaded_weight:
```
you-n-g, line 152: It seems that loading weight makes some assumptions about the loaded data than simply loading the data directly.
It looks like a reusable feature.

Will it be better to create a standalone function for it?

qlib/rl/order_execution/strategy.py
```diff
@@ -268,9 +515,8 @@ def _generate_trade_decision(self, execute_result: list = None) -> BaseTradeDeci
         act = policy_out.act.numpy() if torch.is_tensor(policy_out.act) else policy_out.act
         exec_vols = [self._action_interpreter.interpret(s, a) for s, a in zip(states, act)]
 
-        # In backtest, env.step() needs to be manually called since there is no outer trainer to call it
-        if self._backtest:
-            self._env.step(None)
+        self._state_interpreter.step()
```
you-n-g, line 518: It is weird to help the interpreter maintain its state.
Will it be better to create a `cur_step` attribute in the state returned by the simulator?

qlib/rl/utils/env_wrapper.py
```diff
@@ -196,6 +179,8 @@ def reset(self, **kwargs: Any) -> ObsType:
             )
 
             self.simulator.env = cast(EnvWrapper, weakref.proxy(self))
+            self.state_interpreter.reset()
```
you-n-g, line 182: Please refer to other discussions about interpreter.

tests/rl/test_saoe_simple.py
```diff
@@ -186,25 +186,26 @@ class EmulateEnvWrapper(NamedTuple):
     assert np.sum(obs["data_processed"][60:]) == 0
 
     # second step: action
-    interpreter_action.env = CollectDataEnvWrapper()
-    interpreter_action_twap.env = CollectDataEnvWrapper()
-    interpreter_action.env.reset()
-    interpreter_action_twap.env.reset()
+    interpreter_action.reset()
+    interpreter_action_twap.reset()
```
you-n-g, line 190: Please refer to other discussions about interpreter.

qlib/rl/order_execution/policy.py
lihuoran: Sounds good, but I didn't find a good way to do this. Any suggestions?

qlib/rl/interpreter.py
```diff
@@ -35,12 +31,19 @@ class Interpreter:
     states by calling ``self.env.register_state()``, but it's not planned for first iteration.
     """
 
+    def __init__(self) -> None:
```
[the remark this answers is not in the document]
ultmaster: Agree. I think it's better to have them in EnvWrapperStatus because it's designed for this purpose.
But it would require adding DummyEnv back.

qlib/rl/order_execution/policy.py
you-n-g: @lihuoran 
If I understand it correctly,  it tries to load a policy state dict from a dumped Trainer.
Do you think it is a good idea to create a static utility function for Trainer with a name like `get_policy_state_dict(ckpt_path)`

ultmaster (approved):

PR #1336, opened 2022-11-04
Add customized alpha for customized EMA.

ChiahungTai:
## Description

Add customized alpha for EMA.

## Motivation and Context

Some SW use ema to calculate RSI. But they use alpha=1/N, N=period not spans=N.
See https://github.com/ChiahungTai/MyTT/blob/main/MyTT.py#L150 and https://github.com/ChiahungTai/MyTT/blob/main/MyTT.py#L65.
I also confirm the SW I used for TW stock market use the same formula to calculate RSI.

## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [ ] Fix bugs
- [x] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/data/ops.py
+++ b/qlib/data/ops.py
@@ -1368,8 +1368,9 @@ class EMA(Rolling):
         a feature instance with regression r-value square of given window
     """
 
-    def __init__(self, feature, N):
+    def __init__(self, feature, N, alpha: float = None):
         super(EMA, self).__init__(feature, N, "ema")
+        self.alpha = alpha
 
     def _load_internal(self, instrument, start_index, end_index, *args):
         series = self.feature.load(instrument, start_index, end_index, *args)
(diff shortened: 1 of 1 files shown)


ChiahungTai:
@you-n-g Can you help to review this patch?
Thanks.

qlib/data/ops.py
```diff
@@ -1384,6 +1385,8 @@ def exp_weighted_mean(x):
             series = series.expanding(min_periods=1).apply(exp_weighted_mean, raw=True)
         elif 0 < self.N < 1:
             series = series.ewm(alpha=self.N, min_periods=1).mean()
+        elif self.alpha and 0 < self.alpha < 1:
+            series = series.ewm(alpha=self.alpha, min_periods=1).mean()
```
you-n-g, line 1389: What extra feature does it provide? 
Does the original version provide the same feature if `0 < self.N < 1`?

ChiahungTai: Yes, you are right. I misunderstand what the meaning when the 0< self.N < 1.
I can use the EMA with 1/N when N > 1.
Close the PRs.
Thanks for your feedback. :)

PR #1355, opened 2022-11-14
Fix ZScoreNorm&MinMaxNorm

ChengzhenDu:


ChiahungTai:
You might run black in your local work dir to pass the test.
ex: black . -l 120
you can create a new branch and fix the patch in the same PR.
It is not necessary to create new PR for every modification.

you-n-g:
Thanks for your contribution.
The issue is fixed in https://github.com/microsoft/qlib/pull/1398

Issue #1196, opened 2022-07-13
AttributeError: module 'collector' has no attribute 'YahooNormalizeUS1dExtend' while running update_data_to_bin


rmallof:
## 🐛 Bug Description

After downloading all the US tickers I attempted to update the prices using 'python collector.py update_data_to_bin' 

An observation made is it took almost the same amount of time than downloading all the data. Plus, by the end of the execution, the script crashed:

2022-07-13 06:22:32.848 | INFO     | data_collector.base:_collector:197 - error symbol nums: 0
2022-07-13 06:22:32.848 | INFO     | data_collector.base:_collector:198 - current get symbol nums: 12754
2022-07-13 06:22:32.848 | INFO     | data_collector.base:collector_data:211 - 1 finish.
2022-07-13 06:22:32.848 | INFO     | data_collector.base:collector_data:218 - total 12754, error: 0
Traceback (most recent call last):
  File "collector.py", line 1203, in <module>
    fire.Fire(Run)
  File "/opt/conda/envs/qlib/lib/python3.8/site-packages/fire/core.py", line 141, in Fire
    component_trace = _Fire(component, args, parsed_flag_args, context, name)
  File "/opt/conda/envs/qlib/lib/python3.8/site-packages/fire/core.py", line 466, in _Fire
    component, remaining_args = _CallAndUpdateTrace(
  File "/opt/conda/envs/qlib/lib/python3.8/site-packages/fire/core.py", line 681, in _CallAndUpdateTrace
    component = fn(*varargs, **kwargs)
  File "collector.py", line 1178, in update_data_to_bin
    self.normalize_data_1d_extend(qlib_data_1d_dir)
  File "collector.py", line 1058, in normalize_data_1d_extend
    _class = getattr(self._cur_module, f"{self.normalize_class_name}Extend")
AttributeError: module 'collector' has no attribute 'YahooNormalizeUS1dExtend'

## To Reproduce

1. Download all prices from US region
2. Run python collector.py update_data_to_bin --source_dir /us_source --normalize_dir /us_data_normal --qlib_data_1d_dir /us_data --region US --trading_date 2022-07-09 --end_date 2022-07-12

## Expected Behavior

Prices are updated without crashing

## Environment

Linux
x86_64
Linux-5.10.102.1-microsoft-standard-WSL2-x86_64-with-glibc2.10
#1 SMP Wed Mar 2 00:30:59 UTC 2022

[...]

PR #1358, opened 2022-11-15
Resolve issues while running Automatic update of daily frequency data (from yahoo finance) for US region

HyeongminMoon:
## Description
According to README.md, I tried automatic update with US region.

```
python scripts/data_collector/yahoo/collector.py update_data_to_bin --qlib_data_1d_dir <user data dir> --trading_date <start date> --end_date <end date> --region us
```

Current implements are based on CN index, so I entered some issues:
1. There is no method name for US index normalize, YahooNormalizeUS1dExtend 
(#1196) 
-> **resolved** at commit b2a76df
2. While downloading qlib 1d data, exists_qlib_data always return False
-> **Not resolved**. I resolved this by removing "PRN" from the qlib_data/us_index/instruments/all.txt, but I have no idea how can I integrate this in code. Anyway this is not critial issue(no error, just takes ~2min time additionally)
4. While normalizing, some symbols cause error in pd.read_csv because of their names
-> **resolved** at commit a9cb66b
6. While parsing index, it always tries cn_index/collector
-> **resolved** at commit 0be6b99

I am totally ready for accepting your guidance if there is. Thanks.

## Motivation and Context
Motivated by #1196, making automatic update available for US region.
We will be able to use automatic update(with crontab!) for US region after adopting this request.

## How Has This Been Tested?

- [x] Pass the test by running: `python scripts/data_collector/yahoo/collector.py update_data_to_bin --qlib_data_1d_dir <user data dir> --trading_date <start date> --end_date <end date> --region us`.

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/scripts/data_collector/base.py
+++ b/scripts/data_collector/base.py
@@ -296,8 +296,8 @@ class Normalize:
         columns = pd.read_csv(file_path, nrows=0).columns
         df = pd.read_csv(
             file_path,
-            keep_default_na=False,
             dtype={self._symbol_field_name: str},
+            keep_default_na=False,
             na_values={col: symbol_na if col == self._symbol_field_name else default_na for col in columns},
         )
 
--- a/scripts/data_collector/base.py
+++ b/scripts/data_collector/base.py
@@ -290,6 +290,8 @@ class Normalize:
     def _executor(self, file_path: Path):
         file_path = Path(file_path)
 
+        # some symbol_field values such as TRUE, NA are decoded as True(bool), NaN(np.float) by pandas default csv parsing.
+        # manually defines dtype and na_values of the symbol_field.
         default_na = pd._libs.parsers.STR_NA_VALUES
         symbol_na = default_na.copy()
         symbol_na.remove("NA")
--- a/scripts/data_collector/base.py
+++ b/scripts/data_collector/base.py
@@ -289,30 +289,18 @@ class Normalize:
 
     def _executor(self, file_path: Path):
         file_path = Path(file_path)
+
+        default_na = pd._libs.parsers.STR_NA_VALUES
+        symbol_na = default_na.copy()
+        symbol_na.remove("NA")
+        columns = pd.read_csv(file_path, nrows=0).columns
         df = pd.read_csv(
             file_path,
-            dtype={self._symbol_field_name: str},
             keep_default_na=False,
-            na_values=[
-                "",
-                "#N/A",
-                "#N/A N/A",
-                "#NA",
-                "-1.#IND",
-                "-1.#QNAN",
-                "-NaN",
-                "-nan",
-                "1.#IND",
-                "1.#QNAN",
-                "<NA>",
-                "N/A",
-                "NULL",
-                "NaN",
-                "n/a",
-                "nan",
-                "null",
-            ],
+            dtype={self._symbol_field_name: str},
+            na_values={col: symbol_na if col == self._symbol_field_name else default_na for col in columns},
         )
+
         df = self._normalize_obj.normalize(df)
         if df is not None and not df.empty:
             if self._end_date is not None:
--- a/scripts/data_collector/base.py
+++ b/scripts/data_collector/base.py
@@ -289,7 +289,30 @@ class Normalize:
 
     def _executor(self, file_path: Path):
         file_path = Path(file_path)
-        df = pd.read_csv(file_path)
+        df = pd.read_csv(
+            file_path,
+            dtype={self._symbol_field_name: str},
+            keep_default_na=False,
+            na_values=[
+                "",
+                "#N/A",
+                "#N/A N/A",
+                "#NA",
+                "-1.#IND",
+                "-1.#QNAN",
+                "-NaN",
+                "-nan",
+                "1.#IND",
+                "1.#QNAN",
+                "<NA>",
+                "N/A",
+                "NULL",
+                "NaN",
+                "n/a",
+                "nan",
+                "null",
+            ],
+        )
         df = self._normalize_obj.normalize(df)
         if df is not None and not df.empty:
             if self._end_date is not None:
--- a/qlib/utils/__init__.py
+++ b/qlib/utils/__init__.py
@@ -749,7 +749,7 @@ def exists_qlib_data(qlib_dir):
             return False
 
     # check instruments
-    code_names = set(map(lambda x: x.name.lower(), features_dir.iterdir()))
+    code_names = set(map(lambda x: fname_to_code(x.name.lower()), features_dir.iterdir()))
     _instrument = instruments_dir.joinpath("all.txt")
     miss_code = set(pd.read_csv(_instrument, sep="\t", header=None).loc[:, 0].apply(str.lower)) - set(code_names)
     if miss_code and any(map(lambda x: "sht" not in x, miss_code)):
--- a/scripts/data_collector/yahoo/collector.py
+++ b/scripts/data_collector/yahoo/collector.py
@@ -817,6 +817,10 @@ class YahooNormalizeUS1d(YahooNormalizeUS, YahooNormalize1d):
     pass
 
 
+class YahooNormalizeUS1dExtend(YahooNormalizeUS, YahooNormalize1dExtend):
+    pass
+
+
 class YahooNormalizeUS1min(YahooNormalizeUS, YahooNormalize1minOffline):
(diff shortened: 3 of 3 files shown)


you-n-g:
Please merge master to fix the CI error.  @HyeongminMoon

scripts/data_collector/base.py
```diff
@@ -289,7 +289,30 @@ def __init__(
 
     def _executor(self, file_path: Path):
         file_path = Path(file_path)
-        df = pd.read_csv(file_path)
+        df = pd.read_csv(
+            file_path,
+            dtype={self._symbol_field_name: str},
+            keep_default_na=False,
```
SunsetWolf, line 295: Can I know what is the intention of adding the parameters `keep_default_na` and `na_values`?
I guessed that it might be to cope with the presence of special data in the raw data, and to test my guess, I did a test and found that it can handle these special characters even without these two parameters.
Here is my test script:
image

SunsetWolf:
Regarding the second point of your description, my guess is that it is not possible to determine whether qlib_dir exists or not, which causes GetData to be executed every time update_to_bin is run, which doesn't affect anything, but takes some unnecessary time. Is my guess correct?

HyeongminMoon:
> Regarding the second point of your description, my guess is that it is not possible to determine whether qlib_dir exists or not, which causes GetData to be executed every time update_to_bin is run, which doesn't affect anything, but takes some unnecessary time. Is my guess correct?

Correct.

scripts/data_collector/base.py
HyeongminMoon: The problem was the symbol names 'NA' and 'TRUE'.
qlib_bug_NA_TRUE
I could resolve 'TRUE' case by specifying `dtype`, but still 'NA' case didn't work.
So I directly removed 'NA' from pandas's default `na_values`. But I guess this is not a perfect solution. 
Here is a test script for better understanding:
```
import pandas as pd
import io
st = """
NA,1,2,3
TRUE,4,5,6
SYMBOL,N/A,#NA,NULL
"""

# original case
df = pd.read_csv(io.StringIO(st), header=None)
print(df)

# specify dtype: solve 'TRUE'
df2 = pd.read_csv(io.StringIO(st), header=None, dtype={0:str})
print(df2)

# remove 'NA' from default_na_values: solve 'NA'
na_values = pd._libs.parsers.STR_NA_VALUES
na_values.remove('NA')
df3 = pd.read_csv(io.StringIO(st), header=None, keep_default_na=False, na_values=na_values)
print(df3)
```

SunsetWolf:
> > Regarding the second point of your description, my guess is that it is not possible to determine whether qlib_dir exists or not, which causes GetData to be executed every time update_to_bin is run, which doesn't affect anything, but takes some unnecessary time. Is my guess correct?
> 
> Correct.

If that's the case, I think changing the `not` in [this judgment statement](https://github.com/microsoft/qlib/blob/main/scripts/data_collector/yahoo/collector.py#L1165) might solve the problem.

HyeongminMoon:
> > > Regarding the second point of your description, my guess is that it is not possible to determine whether qlib_dir exists or not, which causes GetData to be executed every time update_to_bin is run, which doesn't affect anything, but takes some unnecessary time. Is my guess correct?
> > 
> > 
> > Correct.
> 
> If that's the case, I think changing the `not` in [this judgment statement](https://github.com/microsoft/qlib/blob/main/scripts/data_collector/yahoo/collector.py#L1165) might solve the problem.

No, I think this is not from `exists_qlib_data`, but from a bug during the download sessions. I found there is a symbol name 'PRN' in the `qlib_data/us_index/instruments/all.txt` while PRN.csv does not exist in downloaded source directory.

SunsetWolf:
> > > > Regarding the second point of your description, my guess is that it is not possible to determine whether qlib_dir exists or not, which causes GetData to be executed every time update_to_bin is run, which doesn't affect anything, but takes some unnecessary time. Is my guess correct?
> > > 
> > > 
> > > Correct.
> > 
> > 
> > If that's the case, I think changing the `not` in [this judgment statement](https://github.com/microsoft/qlib/blob/main/scripts/data_collector/yahoo/collector.py#L1165) might solve the problem.
> 
> No, I think this is not from `exists_qlib_data`, but from a bug during the download sessions. I found there is a symbol name 'PRN' in the `qlib_data/us_index/instruments/all.txt` while PRN.csv does not exist in downloaded source directory.

Since windows cannot create some special name folders, "prn" is one of them, in order to ensure that users of various systems can also use qlib, we add "\_qlib\_" prefix to these special names, so the "prn" seen in `instruments/all.txt`, and `features/_qlib_prn/` are matched , because in `exists_qlib_data`, there is no remove prefix for the names in features, so returns False, we provide [a method to remove the prefix](https://github.com/microsoft/qlib/blob/main/qlib/utils/__init__.py#L1036), I think this method can be used in [the exists_qlib_data code](https://github.com/microsoft/qlib/blob/main/qlib/utils/__init__.py#L752), to solve the problem.

scripts/data_collector/base.py
SunsetWolf: With your reply, I understand the meaning of `keep_default_na` and `na_values` parameters, thanks! But this way of writing makes it difficult to understand the code, I tried a new way of rendering, which makes it easier for the user to understand the code.
Here is my test script:
```
import pandas as pd
import io

st = """
NA,1,2,3
TRUE,4,5,6
SYMBOL,N/A,#NA,NULL
"""

na_cand = [
    "", "#N/A", "#N/A N/A", "#NA", "-1.#IND", "-1.#QNAN", 
    "-NaN", "-nan", "1.#IND", "1.#QNAN", "<NA>", "N/A", 
    "NULL", "NaN", "n/a", "nan", "null",
]

df = pd.read_csv(
    io.StringIO(st),
    header=None,
    dtype={0: str},
    keep_default_na=False,
    na_values={
        0: na_cand,
        1: pd._libs.parsers.STR_NA_VALUES,
        2: pd._libs.parsers.STR_NA_VALUES,
        3: pd._libs.parsers.STR_NA_VALUES,
    }
)
print(df)
```

HyeongminMoon: Thank you for better rendering suggestion! I'll refer to it and commit.

HyeongminMoon:
> > > > > Regarding the second point of your description, my guess is that it is not possible to determine whether qlib_dir exists or not, which causes GetData to be executed every time update_to_bin is run, which doesn't affect anything, but takes some unnecessary time. Is my guess correct?
> > > > 
> > > > 
> > > > Correct.
> > > 
> > > 
> > > If that's the case, I think changing the `not` in [this judgment statement](https://github.com/microsoft/qlib/blob/main/scripts/data_collector/yahoo/collector.py#L1165) might solve the problem.
> > 
> > 
> > No, I think this is not from `exists_qlib_data`, but from a bug during the download sessions. I found there is a symbol name 'PRN' in the `qlib_data/us_index/instruments/all.txt` while PRN.csv does not exist in downloaded source directory.
> 
> Since windows cannot create some special name folders, "prn" is one of them, in order to ensure that users of various systems can also use qlib, we add "_qlib_" prefix to these special names, so the "prn" seen in `instruments/all.txt`, and `features/_qlib_prn/` are matched , because in `exists_qlib_data`, there is no remove prefix for the names in features, so returns False, we provide [a method to remove the prefix](https://github.com/microsoft/qlib/blob/main/qlib/utils/__init__.py#L1036), I think this method can be used in [the exists_qlib_data code](https://github.com/microsoft/qlib/blob/main/qlib/utils/__init__.py#L752), to solve the problem.

I'll try this too. Thanks.

HyeongminMoon:
Now all issues resolved, I checked the script once again too.

scripts/data_collector/base.py
```diff
@@ -289,7 +289,18 @@ def __init__(
 
     def _executor(self, file_path: Path):
         file_path = Path(file_path)
-        df = pd.read_csv(file_path)
+
+        default_na = pd._libs.parsers.STR_NA_VALUES
```
you-n-g, line 293: It would be great if we could add more context and comments here.
It is hard to understand the code without an explanation of motivation.

HyeongminMoon: Thank you for review! I added explanation comments:)

you-n-g:
It looks perfect now.
Thanks!

PR #1364, opened 2022-11-17
Optimize the implementation of uri & Fix async log bug

you-n-g:
## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [ ] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/workflow/__init__.py
+++ b/qlib/workflow/__init__.py
@@ -8,7 +8,6 @@ from .exp import Experiment
 from .recorder import Recorder
 from ..utils import Wrapper
 from ..utils.exceptions import RecorderInitializationError
-from qlib.config import C
 
 
 class QlibRecorder:
@@ -347,14 +346,14 @@ class QlibRecorder:
 
     def set_uri(self, uri: Optional[Text]):
         """
-        Method to reset the current uri of current experiment manager.
+        Method to reset the **default** uri of current experiment manager.
 
         NOTE:
 
         - When the uri is refer to a file path, please using the absolute path instead of strings like "~/mlruns/"
           The backend don't support strings like this.
         """
-        self.exp_manager.set_uri(uri)
+        self.exp_manager.default_uri = uri
 
     @contextmanager
     def uri_context(self, uri: Text):
@@ -370,11 +369,11 @@ class QlibRecorder:
             the temporal uri
         """
         prev_uri = self.exp_manager.default_uri
-        C.exp_manager["kwargs"]["uri"] = uri
+        self.exp_manager.default_uri = uri
         try:
             yield
         finally:
-            C.exp_manager["kwargs"]["uri"] = prev_uri
+            self.exp_manager.default_uri = prev_uri
 
     def get_recorder(
         self,
--- a/qlib/workflow/expm.py
+++ b/qlib/workflow/expm.py
@@ -24,7 +24,7 @@ class ExpManager:
         This is the `ExpManager` class for managing experiments. The API is designed similar to mlflow.
         (The link: https://mlflow.org/docs/latest/python_api/mlflow.html)
 
-        The `ExpManager` is expected to be a singleton (btw, we can have multiple `Experiment`s with different uri. user can get different experiments from different uri, and then compare records of them). Global Config (e.g. `C`)  is also a singleton.
+        The `ExpManager` is expected to be a singleton (btw, we can have multiple `Experiment`s with different uri. user can get different experiments from different uri, and then compare records of them). Global Config (i.e. `C`)  is also a singleton.
     So we try to align them together.  They share the same variable, which is called **default uri**. Please refer to `ExpManager.default_uri` for details of variable sharing.
 
         When the user starts an experiment, the user may want to set the uri to a specific uri (it will override **default uri** during this period), and then unset the **specific uri** and fallback to the **default uri**.    `ExpManager._active_exp_uri` is that **specific uri**.
@@ -317,9 +317,6 @@ class MLflowExpManager(ExpManager):
     """
     Use mlflow to implement ExpManager.
     """
-
-    def __init__(self, uri: Text, default_exp_name: Optional[Text]):
-        super(MLflowExpManager, self).__init__(uri, default_exp_name)
 
     @property
     def client(self):
--- a/qlib/workflow/expm.py
+++ b/qlib/workflow/expm.py
@@ -60,7 +60,7 @@ class ExpManager:
         Start an experiment. This method includes first get_or_create an experiment, and then
         set it to be active.
 
-        Maintaining `_active_exp_uri` is included in start_exp, remaining implementation should be implemented in _end_exp in subclass
+        Maintaining `_active_exp_uri` is included in start_exp, remaining implementation should be included in _end_exp in subclass
 
         Parameters
         ----------
@@ -100,7 +100,7 @@ class ExpManager:
         """
         End an active experiment.
 
-        Maintaining `_active_exp_uri` is included in end_exp, remaining implementation should be implemented in _end_exp in subclass
+        Maintaining `_active_exp_uri` is included in end_exp, remaining implementation should be included in _end_exp in subclass
 
         Parameters
         ----------
@@ -302,16 +302,6 @@ class ExpManager:
         """
         return self._active_exp_uri or self.default_uri
 
-    def set_uri(self, uri: Text):
-        """
-        Set the default tracking URI and the corresponding variables.
-
-        Parameters
-        ----------
-        uri  : str
-        """
-        self.default_uri = uri
-
     def list_experiments(self):
         """
         List all the existing experiments.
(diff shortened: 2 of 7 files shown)


you-n-g:
@ChiahungTai  You are welcome to review this PR.

qlib/workflow/expm.py
```diff
@@ -15,23 +15,35 @@
...
 
+
 logger = get_module_logger("workflow")
 
 
 class ExpManager:
     """
-    This is the `ExpManager` class for managing experiments. The API is designed similar to mlflow.
-    (The link: https://mlflow.org/docs/latest/python_api/mlflow.html)
+        This is the `ExpManager` class for managing experiments. The API is designed similar to mlflow.
```
you-n-g, line 24: @ChiahungTai  You can start from these docs.

qlib/workflow/expm.py
```diff
@@ -15,23 +15,35 @@
...
+        When the user starts an experiment, the user may want to set the uri to a specific uri (it will override **default uri** during this period), and then unset the **specific uri** and fallback to the **default uri**.    `ExpManager._active_exp_uri` is that **specific uri**.
     """
 
+    active_experiment: Optional[Experiment]
+
     def __init__(self, uri: Text, default_exp_name: Optional[Text]):
-        self._current_uri = uri
+        # The `ExpManager` is expected to be a singleton.
+        # So we just use `C.exp_manager["kwargs"]["uri"]` for `uri` instead.
+        # Setting the variable to Config will make the uri consistant for multi-processing in all platforms
```
you-n-g, line 38:
```suggestion
```

ChiahungTai:
Fail the test in test_all_pipeline.py.
I have add the test case for you in #1363.

qlib/workflow/expm.py
ChiahungTai: "we can have multiple `Experiment`s with different uri" => IIUC, the words are imply the design of ExpManger is suggestion the user use different uri in each experiment? 

From my understanding in MLFLow, the uri is like a store backend(sql, file...). The ExpManger is a single entry to lookup and manipulate the experiments in the uri.
 
So the user can change different uri to retrieve the different topic of experiements.

ChiahungTai:
Looks like the exception is the same as the test case #1363.

            except MlflowException as e:
                raise ValueError(
                    "No valid experiment has been found, please make sure the input experiment name is correct."
                ) from e

qlib/workflow/__init__.py
```diff
@@ -370,11 +369,11 @@ def uri_context(self, uri: Text):
             the temporal uri
         """
         prev_uri = self.exp_manager.default_uri
-        C.exp_manager["kwargs"]["uri"] = uri
+        self.exp_manager.default_uri = uri
```
ChiahungTai, line 372: I think the set of default_uri is not engouh. The self.exp_manager is not change the client uri if you only change default uri.

qlib/workflow/expm.py
```diff
@@ -345,12 +355,11 @@ def start_exp(
...
         return self.active_experiment
 
-    def end_exp(self, recorder_status: Text = Recorder.STATUS_S):
+    def _end_exp(self, recorder_status: Text = Recorder.STATUS_S):
         if self.active_experiment is not None:
             self.active_experiment.end(recorder_status)
             self.active_experiment = None
-        # When an experiment end, we will release the current uri.
-        self._current_uri = None
+        self._set_client_uri()
```
ChiahungTai, line 362: Pump into an exception - No argument.

qlib/workflow/expm.py
```diff
@@ -322,18 +333,17 @@ def client(self):
             self._client = mlflow.tracking.MlflowClient(tracking_uri=self.uri)
         return self._client
```
ChiahungTai, line 334: I think you might need to sync the returned self._client to the updated default_uri. 
Or you have set the uri in every function in access the client.

qlib/workflow/expm.py
you-n-g: Yes.  
ExpManager is a singleton, but we can change its uri  to get experiments from different uri.
I added some comments just now.
Please check if it is helpful

qlib/workflow/recorder.py
```diff
@@ -329,7 +329,7 @@ def get_local_dir(self):
     def start_run(self):
         # set the tracking uri
         mlflow.set_tracking_uri(self.uri)
-        # start the run
+        # start the RuntimeError
```
ChiahungTai, line 332: What does the start the RuntimeError mean?

tests/dependency_tests/test_mlflow.py
```diff
@@ -0,0 +1,34 @@
...
+        we don't cache _client (this is helpful to reduce maintainance work when MLflowExpManager's uri is chagned)
+
+        This implementation is based on the assumption creating a client is fast
+        """
+        start = time.time()
+        for i in range(10):
+            _ = mlflow.tracking.MlflowClient(tracking_uri=str(self.TMP_PATH))
+        end = time.time()
+        elasped = end - start
+        self.assertGreater(1e-2, elasped)  # it can be done in less than 10ms
```
ChiahungTai, line 29: assertGreater or assertLesser

qlib/workflow/expm.py
ChiahungTai: Your new comment is more clear than before.

qlib/workflow/__init__.py
you-n-g: It is fixed now.
The _client will not be cached.  No further maintenance is required for it.

qlib/workflow/expm.py
you-n-g: Fixed Now.

tests/dependency_tests/test_mlflow.py
you-n-g: I have changed the direction. And it is more readable than before

you-n-g:
@ChiahungTai 
Hi, I have fixed all the comments and waiting for the CI.

I checked out [your test](https://github.com/microsoft/qlib/pull/1363/files#diff-e85369ca90d47fb4376b973050e0f2bd66cca8aa16ab93949415c3af58b2a288R85)

I didn't get what you want to assert in the test.

It seems that the exception looks reasonable.
When we want to get a recorder,  Qlib tries to get an experiment first.   The experiment name or id is not given.
So the default exp name is used. 
There is no experiment which is active or is named with the default exp name.
So an exception of `No valid experiment has been found` is raised

qlib/workflow/recorder.py
you-n-g: Fixed now

qlib/workflow/expm.py
you-n-g: Fixed now.

ChiahungTai:
The test code is pass after assign the experiment_name and recorder_name.
But in my own project, I provide the experiment_name and recorder_name.
But the code is too long to write a simple test case.
`
    with R.uri_context(uri=uri_path):
        recorder = R.get_recorder(experiment_name=EXP_NAME, recorder_name="original")
`
I am working on another version of cached_pipeline in my own project. I might test the new commit and try it my local first.
Also I will upload a new test if I can rewrite a simple test case.

> @ChiahungTai Hi, I have fixed all the comments and waiting for the CI.
> 
> I checked out [your test](https://github.com/microsoft/qlib/pull/1363/files#diff-e85369ca90d47fb4376b973050e0f2bd66cca8aa16ab93949415c3af58b2a288R85)
> 
> I didn't get what you want to assert in the test.
> 
> It seems that the exception looks reasonable. When we want to get a recorder, Qlib tries to get an experiment first. The experiment name or id is not given. So the default exp name is used. There is no experiment which is active or is named with the default exp name. So an exception of `No valid experiment has been found` is raised

ChiahungTai:
Your new patch works fine in my own project.

> The test code is pass after assign the experiment_name and recorder_name. But in my own project, I provide the experiment_name and recorder_name. But the code is too long to write a simple test case. `with R.uri_context(uri=uri_path): recorder = R.get_recorder(experiment_name=EXP_NAME, recorder_name="original")` I am working on another version of cached_pipeline in my own project. I might test the new commit and try it my local first. Also I will upload a new test if I can rewrite a simple test case.
> 
> > @ChiahungTai Hi, I have fixed all the comments and waiting for the CI.
> > I checked out [your test](https://github.com/microsoft/qlib/pull/1363/files#diff-e85369ca90d47fb4376b973050e0f2bd66cca8aa16ab93949415c3af58b2a288R85)
> > I didn't get what you want to assert in the test.
> > It seems that the exception looks reasonable. When we want to get a recorder, Qlib tries to get an experiment first. The experiment name or id is not given. So the default exp name is used. There is no experiment which is active or is named with the default exp name. So an exception of `No valid experiment has been found` is raised

ChiahungTai:
The patch LGTM. You can merge it after fix the pylint error.

PR #1407, opened 2023-01-03
[BUGFIX] allow sell in limit-up case and allow buy in limit-down case in topk strategy

qianyun210603:
## Description
Allow sell in limit-up case and allow buy in limit-down case
Also fixed some typo/typehints.

## Motivation and Context
To be consistent with real world, in which we can buy at limit-down (跌停) and sell at limit up (涨停).

## How Has This Been Tested?

1. Pipeline test:
image
2. Your own tests:

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation


qlib/contrib/strategy/signal_strategy.py
```diff
@@ -216,7 +217,7 @@ def filter_stock(li):
         buy = today[: len(sell) + self.topk - len(last)]
         for code in current_stock_list:
             if not self.trade_exchange.is_stock_tradable(
-                stock_id=code, start_time=trade_start_time, end_time=trade_end_time
+                stock_id=code, start_time=trade_start_time, end_time=trade_end_time, direction=OrderDir.SELL
```
you-n-g, line 220: Thanks for your contribution.  @qianyun210603 
Not selling limit-up stocks and buying limit-down could be considered as a part of the strategy in A-stock.
Although current implementation seems a little weird,  but it is a reasonable results.

Your version is better. But if we merge it currently, all the previous results should be updated to make the code and results consistent (it will take a long time to update all the results).

Would it be better if we add it as a feature (a parameter in the init method) and keep the previous version as the default behavior? 
(Besides, TODOs to switch to your version should  be added in the docs/docstrings)

qianyun210603: Added a flag in `TopkDropoutStrategy` defaults to `True` to keep previous behaviour as suggested.

you-n-g:
Thanks!  It looks great now!

PR #1410, opened 2023-01-06
Update data.py

wony-zheng:
to suppress this warning:
FutureWarning: The default dtype for empty Series will be 'object' instead of 'float64' in a future version. Specify a dtype explicitly to silence this warning.
  return pd.Series()


you-n-g:
Could you please write a test to reproduce this error?
Thanks.

PR #1413, opened 2023-01-09
Enhancement and bugfix in report plots

qianyun210603:
## Description
1. allow input custom benchmark return series to compare with group1.
2. fix stratifying bug: simply using `len(x) // N * i : len(x) // N * (i + 1)` misses `len(x) % N` names at end.
3. force colorbar in monthly ic heat map centers at 0.

## Motivation and Context
1. more convenient to group returns with index.
2. bug fix.
3. consider monthly ic series [0.1, -0.1, 0.1, -0.1, 0, 0.1] and [0.3, 0.1, 0.3, 0.1, 0.2, 0.3]. Apparently the former is bad strategy with unstable ic and the latter is good strategy, but if not forcing symmetric color bar, the two gives same heatmap look (except numbers in colour axis), which is not good visualisation. 

## How Has This Been Tested?
Generated plot locally.

1. Pipeline test:
5. Your own tests:
image

## Types of changes

- [x] Fix bugs and enhancements
- [ ] Add new feature
- [ ] Update documentation


qianyun210603:
@you-n-g would you review please?

qlib/contrib/report/analysis_model/analysis_model_performance.py
```diff
@@ -49,8 +74,17 @@ def _group_return(pred_label: pd.DataFrame = None, reverse: bool = False, N: int
     # Long-Short
     t_df["long-short"] = t_df["Group1"] - t_df["Group%d" % N]
 
-    # Long-Average
-    t_df["long-average"] = t_df["Group1"] - pred_label.groupby(level="datetime")["label"].mean()
+    # Long-benchmark
+    benchmark = kwargs.get("benchmark", "average")
```
Fivele-Li, line 78: I think it would be better to have the benchmark as an explicit input parameter, and we can also add type hints.

qlib/contrib/report/analysis_model/analysis_model_performance.py
```diff
@@ -49,8 +74,17 @@ def _group_return(pred_label: pd.DataFrame = None, reverse: bool = False, N: int
...
+    if isinstance(benchmark, str) and benchmark == "average":
+        benchmark_name = benchmark
+        benchmark = pred_label.groupby(level="datetime")["label"].mean()
+    elif isinstance(benchmark, pd.Series):
+        benchmark_name = benchmark.name if bool(benchmark.name) else "benchmark"
+        benchmark = benchmark.reindex(t_df.index)
```
Fivele-Li, line 84: There is a high possibility that the user-provided benchmark may not align with t_df.index. Perhaps we can add some common warnings in this case if the DataFrame doesn't have warning messages.

qianyun210603: I think it might not be necessary as it should be a common case especially for a global portfolio. Think about using CSI300 to benchmarking a portfolio which invests both A and H shares, for those mainland holidays which are not HK holidays.

qlib/contrib/report/analysis_model/analysis_model_performance.py
qianyun210603: changed as suggested.

qianyun210603:
any further comments?

PR #1417, opened 2023-01-18
RL backtest pipeline on 5-min data

lihuoran:
## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [ ] Fix bugs
- [ ] Add new feature
- [ ] Update documentation


qlib/rl/order_execution/strategy.py
```diff
@@ -89,6 +89,7 @@ def __init__(
         exchange: Exchange,
         ticks_per_step: int,
         backtest_data: IntradayBacktestData,
+        data_granularity: int = 1,
```
ultmaster, line 92: Maybe data_granularity should be a property of IntraDayBacktestData?

qlib/rl/order_execution/reward.py
```diff
@@ -47,3 +48,51 @@ def reward(self, simulator_state: SAOEState) -> float:
...
+            intraday_data = simulator_state.backtest_data.data[self.start_time_index:self.end_time_index + 1]
+            
+            if traded >= 1e-6:
+                vwap_price = trade_value / traded
+            else:
+                vwap_price = _weighted_average(  # VWAP price of the entire day
+                    val=intraday_data["$close0"].to_numpy(), 
+                    weight=intraday_data["$volume0"].to_numpy(),
+                )
+            twap_price = np.mean(intraday_data["$close0"].to_numpy())
```
Arthur-Null, line 88: We should use actual deal price instead of $close0. Might use
```
vwap_price = simulator_state.metrics.trade_price
twap_price = simulator_state.backtest_data.get_deal_price().mean()
```

Arthur-Null: Also, we need to add the judgement for twap = 0.

rk2900, line 88: may need a judgement for voiding dividing zero as calculated below.

you-n-g:
Please fix the CI error
image

qlib/rl/order_execution/strategy.py
lihuoran: Personally think it is better to leave it as it is after several attempts. We can fix this in the future if we find a better design.

PR #1463, opened 2023-03-15
fix_pylint_w0719

SunsetWolf:
## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [ ] Fix bugs
- [ ] Add new feature
- [ ] Update documentation


SunsetWolf:
For the changes in the `recorder.py` file, I think it is better to raise `ValueError` because `self._artifact_uri` is defined during initialization; if `self._artifact_uri` is not defined, it is more appropriate to raise `AttributeError`.

PR #1481, opened 2023-03-30
Add exploration noise to rl training collector

chenditc:
Add exploration_noise=True  to training collector.

## Description

DQN family need to use this parameter to randomly sample action.

Default behavior won't change action, this will only affect policy which implemented exploration_noise function : 
https://github.com/thu-ml/tianshou/blob/7f8fa241dd501dbab7dba5c7e95f66702ecf039b/tianshou/policy/base.py#L92

## Motivation and Context

Current rl framework don't support DQN family algorithm

## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [ ] Fix bugs
- [x] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/rl/trainer/vessel.py
+++ b/qlib/rl/trainer/vessel.py
@@ -168,7 +168,7 @@ class TrainingVessel(TrainingVesselBase):
         self.policy.train()
 
         with vector_env.collector_guard():
-            collector = Collector(self.policy, vector_env, VectorReplayBuffer(self.buffer_size, len(vector_env)))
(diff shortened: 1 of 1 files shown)


Fivele-Li:
I think it would be better to make the exploration_noise a configurable variable instead of hardcoding it in the code.

chenditc:
This is in the train function, I don't see any scenario where we want to disable exploration when training a RL agent.

Fivele-Li:
Thank you so much for taking the time to submit your pull request. Your contribution is greatly appreciated!

PR #1509, opened 2023-05-10
Add util function to help automatically get horizon

chenditc:
Add util function to help automatically get horizon

## Description
Add util function to help automatically get horizon

## Motivation and Context

In some code like TRA and DDG-DA, horizon is manually passed into the model. Which is not convenient to automate the workflow. Add a util function to help guess the horizon by label.

## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [ ] Fix bugs
- [x] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/utils/data.py
+++ b/qlib/utils/data.py
@@ -3,10 +3,12 @@
 """
 This module covers some utility functions that operate on data or basic object
 """
+import re
 from copy import deepcopy
 from typing import List, Union
-import pandas as pd
+
 import numpy as np
+import pandas as pd
 
 
 def robust_zscore(x: pd.Series, zscore=False):
@@ -103,3 +105,19 @@ def update_config(base_config: dict, ext_config: Union[dict, List[dict]]):
                     # one of then are not dict. Then replace
                     base_config[key] = ec[key]
     return base_config
+
+
+def guess_horizon(label):
+    """
+        Try to guess the horizon by parsing label
+    """
+    regex = r"Ref\(\s*\$[a-zA-Z]+,\s*-(\d+)\)"
+    horizon_list = [int(x) for x in re.findall(regex, label)]
+
+    if len(horizon_list) == 0:
+        return None
+    max_horizon = max(horizon_list)
+    # Unlikely the label doesn't use future information
+    if max_horizon < 2:
+        return None
+    return max_horizon
--- a/tests/misc/test_utils.py
+++ b/tests/misc/test_utils.py
@@ -9,6 +9,7 @@ from qlib.config import C
 from qlib.log import TimeInspector
 from qlib.constant import REG_CN, REG_US, REG_TW
 from qlib.utils.time import cal_sam_minute as cal_sam_minute_new, get_min_cal, CN_TIME, US_TIME, TW_TIME
+from qlib.utils.data import guess_horizon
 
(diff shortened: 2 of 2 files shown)


qlib/contrib/data/dataset.py
```diff
@@ -130,6 +132,14 @@ def __init__(
         input_size=None,
         **kwargs,
     ):
+        if horizon == 0:
+            # Try to guess horizon
+            if isinstance(handler, dict):
+                handler = init_instance_by_config(handler)
+            elif isinstance(handler, str):  # pickled handler
```
Fivele-Li, line 139: Could we write it like this?
`if isinstance(handler, (dict, str))`

qlib/utils/data.py
```diff
@@ -103,3 +105,19 @@ def update_config(base_config: dict, ext_config: Union[dict, List[dict]]):
...
+    regex = r"Ref\(\s*\$[a-zA-Z]+,\s*-(\d+)\)"
+    horizon_list = [int(x) for x in re.findall(regex, label)]
+
+    if len(horizon_list) == 0:
+        return 0
+    max_horizon = max(horizon_list)
+    # Unlikely the label doesn't use future information
+    if max_horizon < 2:
+        return 0
+    return max_horizon + 1
```
Fivele-Li, line 123: Why is it `return max_horizon+1` here instead of `return max_horizon` or `return max_horizon-min_horizon`?

chenditc: This is consistent with exisiting deinition of "horizon" in other code like DDG-DA and TRA. There is no official documentation, but it means the number of future days should be truncated to avoid leaking future information. Using max_horizon + 1 just align with exisitng code base.

qlib/contrib/data/dataset.py
chenditc: Done

qlib/utils/data.py
```diff
@@ -103,3 +105,19 @@ def update_config(base_config: dict, ext_config: Union[dict, List[dict]]):
                     # one of then are not dict. Then replace
                     base_config[key] = ec[key]
     return base_config
+
+
+def guess_horizon(label):
```
you-n-g, line 110: Should this function be removed now?

qlib/contrib/data/dataset.py
```diff
@@ -130,6 +132,12 @@ def __init__(
         input_size=None,
         **kwargs,
     ):
+        if horizon == 0:
```
you-n-g, line 135: Assert we have label  field  in the dataset.

qlib/utils/data.py
```diff
@@ -103,3 +105,19 @@ def update_config(base_config: dict, ext_config: Union[dict, List[dict]]):
...
+    """
+    Try to guess the horizon by parsing label
+    """
+    regex = r"Ref\(\s*\$[a-zA-Z]+,\s*-(\d+)\)"
```
you-n-g, line 114: https://github.com/microsoft/qlib/blob/main/qlib/data/data.py#L543

you-n-g: https://github.com/microsoft/qlib/blob/main/qlib/data/data.py#L853C44-L853C68

you-n-g: rght_etd

PR #1546, opened 2023-06-10
Add multi pass portfolio analysis record

chenditc:
## Description
Add a new record to initialize position randomly and run backtest multiple times.

## Motivation and Context
Signal based strategy's performance can be variant due to initial position setup. To better evalute the performance of each model, this recorder will:
1. Randomly shuffle the prediction signal of first backtest date, so that initial position will be random.
2. Run the backtest multiple times.
3. Aggregate the portfolio performance and publish the metrics.

## How Has This Been Tested?

Added an example benchmark config yaml, run with the example yaml. 
```
$ qrun workflow_config_backward_alstm_Alpha158WithBenchDiff_full.yaml
```

1. Pipeline test:
image

4. Your own tests:
image

I also ran test on existing workflow file and confirmed the format of out is the same.

## Types of changes

- [ ] Fix bugs
- [x] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/workflow/record_temp.py
+++ b/qlib/workflow/record_temp.py
@@ -7,7 +7,7 @@ import pandas as pd
 import numpy as np
 from tqdm import trange
 from pprint import pprint
-from typing import Union, List, Optional
+from typing import Union, List, Optional, Dict
 
 from qlib.utils.exceptions import LoadObjectError
 from ..contrib.evaluate import risk_analysis, indicator_analysis
@@ -237,7 +237,7 @@ class ACRecordTemp(RecordTemp):
             self.save(**artifact_dict)
         return artifact_dict
 
-    def _generate(self, *args, **kwargs) -> dict[str, object]:
+    def _generate(self, *args, **kwargs) -> Dict[str, object]:
         """
         Run the concrete generating task, return the dictionary of the generated results.
         The caller method will save the results to the recorder.
--- a/qlib/workflow/record_temp.py
+++ b/qlib/workflow/record_temp.py
@@ -19,6 +19,7 @@ from ..log import get_module_logger
 from ..utils import fill_placeholder, flatten_dict, class_casting, get_date_by_shift
 from ..utils.time import Freq
 from ..utils.data import deepcopy_basic_type
+from ..utils.exceptions import QlibException
 from ..contrib.eva.alpha import calc_ic, calc_long_short_return, calc_long_short_prec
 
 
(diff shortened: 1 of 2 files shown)


qlib/workflow/record_temp.py
```diff
@@ -554,3 +554,121 @@ def list(self):
...
+            first_bt_pred_date = all_pred_dates[all_pred_dates >= bt_start_date].min()
+
+        # Shuffle the first backtest date's pred score
+        first_date_score = pred_df.loc[first_bt_pred_date]["score"]
+        np.random.shuffle(first_date_score.values)
+
+        # Reset strategy so that pred df can be replaced in next generate
+        self.strategy_config = deepcopy_basic_type(self.original_strategy)
+
+        self.save(**{"pred.pkl": pred_df})
```
you-n-g, line 616: Why do we need to save the shuffled prediction?

qlib/workflow/record_temp.py
```diff
@@ -554,3 +554,121 @@ def list(self):
...
+    def _generate(self, **kwargs):
+        risk_analysis_df_map = {}
+
+        # Collect each frequency's analysis df as df list
+        for i in range(self.pass_num):
+            if self.shuffle_init_score:
+                self.random_init()
+
+            # Not check for cache file list
+            super()._generate(**kwargs)
```
you-n-g, line 627: It is generating files that is out of the scope of the class. 
Would you think it is better to redefine  ` ACRecordTemp  _generate` to `def _generate(self, *args, **kwargs) -> Dict[str, obj]:`
And save the return objects in ` ACRecordTemp  generate`?

qlib/workflow/record_temp.py
```diff
@@ -554,3 +554,121 @@ def list(self):
...
+        # Reset strategy so that pred df can be replaced in next generate
+        self.strategy_config = deepcopy_basic_type(self.original_strategy)
+
+        self.save(**{"pred.pkl": pred_df})
+
+    def _generate(self, **kwargs):
+        risk_analysis_df_map = {}
+
+        # Collect each frequency's analysis df as df list
+        for i in range(self.pass_num):
```
you-n-g, line 622: I think adding tqdm would  improve the UE.

qlib/workflow/record_temp.py
you-n-g: Besides , this will make other code simpler

qlib/workflow/record_temp.py
chenditc: Done

qlib/workflow/record_temp.py
chenditc: Done

chenditc:
The pipeline failure seems related to global python setup issue, not caused by my change.

you-n-g:
Yes. It is due to an error in data storage. We'll fix it later.

chenditc:
> @chenditc The CI issue on the master branch has been fixed. Please merge it to run the latest CI.

@you-n-g  CI has passed. Good to merge?

Fivele-Li:
Merged! Thanks a lot~

PR #1581, opened 2023-07-04
fix singleton bug

peteryang1:
## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [ ] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/finco/utils.py
+++ b/qlib/finco/utils.py
@@ -3,14 +3,16 @@ import json
 from fuzzywuzzy import fuzz
 
 
-class Singleton:
+class Singleton(type):
     _instance = None
 
-    def __new__(cls, *args, **kwargs):
+    def __call__(cls, *args, **kwargs):
         if cls._instance is None:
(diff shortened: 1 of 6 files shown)


qlib/finco/utils.py
```diff
@@ -3,14 +3,16 @@
...
     _instance = None
 
-    def __new__(cls, *args, **kwargs):
+    def __call__(cls, *args, **kwargs):
         if cls._instance is None:
-            cls._instance = super().__new__(cls, *args, **kwargs)
+            cls._instance = super(Singleton, cls).__call__(*args, **kwargs)  
         return cls._instance
 
+class SingletonBaseClass(metaclass=Singleton):
```
you-n-g, line 14: I tried [this example](https://github.com/you-n-g/cheatsheets/blob/master/language/python/meta.py#L74)  before.
I can't understand why we need two level of class for Singleton?

Issue #1665, opened 2023-10-13
The error in float calculations


okokkokok:
In Position.py, the function sell_stock() will check np.isclose(self.position[stock_id]["amount"], trade_amount), but I found a bug that this check is False, and self.position[stock_id]["amount"] < -1e-5 is True. The reason of this problem may be the error in float calculations.

PR #1666, opened 2023-10-13
fixamount

okokkokok:
#1665


you-n-g:
Thanks for you pull request.

Current implementation is inconsistent. 
It sell all the remain stock. But the cash is increase based on the original intented amount.

If it is caused by float precision, we may try to solve this by making the error tolerance (i.e. -1e-5) lager.

Abhijais4896 (commented):
if np.isclose(self.position[stock_id]["amount"], trade_amount):

            if np.isclose(self.position[stock_id]["amount"], trade_amount) or self.position[stock_id]["amount"] <= trade_amount:

Issue #1757, opened 2024-03-11
HS_SYMBOLS_URL 404


zhstark:
## 🐛 Bug Description

The URL used in scripts/data_collector/utils.py is dead.

## To Reproduce

1. run collector.py under scripts/data_collector/pit
2. it gets stuck after logging "get cn stock symbols....."
3. debug finds that `requests.get(HS_SYMBOLS_URL.format(s_type=_k), timeout=None)` gets 404 response.

## Expected Behavior

HS_SYMBOLS_URL return right symbols

## Environment

**Note**: User could run `cd scripts && python collect_info.py all` under project directory to get system information
and paste them here directly.

 - Qlib version: "0.9.3.99"
 - Python version: 3.8
 - OS (`Windows`, `Linux`, `MacOS`):MacOS

fhu889:
I encountered the same issue. Excuse me, have you resolved it?

PR #1758, opened 2024-03-12
fix the bug that the HS_SYMBOLS_URL is 404

zhstark:
## Description

Discard using HS_SYMBOLS_URL in `scripts/data_collector/utils.py` `get_hs_stock_symbols()`, since the url is no longer available. 
Use  api of `eastmoney.com/api`, the same api that used in method `get_us_stock_symbols()`, to get hs stock symbols.

## Motivation and Context
it can solve the issue [#1757 ](https://github.com/microsoft/qlib/issues/1757)

## How Has This Been Tested?

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation


zhstark:

SunsetWolf:
Thank you for submitting pull requests, after testing found that the current code can't fetch the data successfully, the code in the collection of `_get_symbol` method should yahoo's specification, but currently it doesn't. Also please fix CI.

PR #1895, opened 2025-03-13
Implement geometric accumulation mode for risk_analysis function (#964)

### LINKED ISSUE #964: Concerns about the calculation of return
## 🐛 Bug Description

I found the calculation of return in backtest confusing. The return is defined on each day's basis [here](https://github.com/microsoft/qlib/blob/2ed806c8469904e441cac494fad2af104a4dbd1d/qlib/backtest/account.py#L260). The return on each day are summed up directly to obtain the annualized return [here](https://github.com/microsoft/qlib/blob/2ed806c8469904e441cac494fad2af104a4dbd1d/qlib/contrib/evaluate.py#L58). Shouldn't the total return be a product of (1 + r_i) minus 1? What's the problem here?

## To Reproduce

1.
1.
1.

## Environment

and paste them here directly.

 - Qlib version: 2ed806c8469904e441cac494fad2af104a4dbd1d
 - Python version: 3.7
 - OS (`Windows`, `Linux`, `MacOS`): Linux
 - Commit number (optional, please provide it if you are using the dev version): 2ed806c8469904e441cac494fad2af104a4dbd1d

eabjab:
## Description
This change adds a new parameter `mode` to the `risk_analysis` function which determines the accumulation method to be used when calculating returns. Default parameter value for `mode` is "sum", which maintains original behavior. 

If `mode` is "product", the following calculations are used instead of the current arithmetic methods to align with expectations for geometrically accumulated returns:

`mean = np.prod(1 + r) ** (1 / len(r)) - 1    # geometric mean`
`std = np.log(1 + r).std(ddof=1)    # volatility of log returns`
`cumulative_return = np.prod(1 + r) - 1`
`annualized_return = (1 + cumulative_return) ** (N / len(r)) - 1`
`max_drawdown = ((1 + r).cumprod() / (1 + r).cumprod().cummax() - 1).min()    # max percentage drawdown from peak cumulative product`
`information_ratio` uses the same equation in "product" mode as in "sum" mode, but with geometric mean and volatility of log returns instead of their arithmetic counterparts.

These calculations are what I would intuitively expect for geometrically compounded returns (there weren't too many specifics in the issue thread), but I'm happy to change any of the underlying calculations if a different equation should be used for any of the metrics in "product" mode.

## Motivation and Context
Resolves #964 as well as the TODO inside `risk_analysis`

Adding a geometric/product accumulation method to the `risk_analysis` function is a new feature that aligns with what many users would expect in quant finance.

## How Has This Been Tested?

1. Pipeline test: 
qlib_issue#964_pipeline_tests

2. Your own tests:
qlib_issue#964_custom_tests

## Types of changes

- [ ] Fix bugs
- [x] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/contrib/evaluate.py
+++ b/qlib/contrib/evaluate.py
@@ -7,7 +7,7 @@ from __future__ import print_function
 import numpy as np
 import pandas as pd
 import warnings
-from typing import Union
+from typing import Union, Literal
 
 from ..log import get_module_logger
 from ..utils import get_date_range
@@ -24,15 +24,13 @@ from ..data.dataset.utils import get_level_index
 logger = get_module_logger("Evaluate")
 
 
-def risk_analysis(r, N: int = None, freq: str = "day"):
+def risk_analysis(r, N: int = None, freq: str = "day", mode: Literal["sum", "product"] = "sum"):
     """Risk Analysis
     NOTE:
-    The calculation of annulaized return is different from the definition of annualized return.
+    The calculation of annualized return is different from the definition of annualized return.
     It is implemented by design.
-    Qlib tries to cumulated returns by summation instead of production to avoid the cumulated curve being skewed exponentially.
+    Qlib tries to cumulate returns by summation instead of production to avoid the cumulated curve being skewed exponentially.
     All the calculation of annualized returns follows this principle in Qlib.
-
-    TODO: add a parameter to enable calculating metrics with production accumulation of return.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


eabjab:

SunsetWolf:
Hi, @eabjab 
Please let the CI pass by pulling the main branch.

SunsetWolf:
Hi, @eabjab 
It was nice to see your PR, and I'm glad to see it. But since you didn't reply. So I created a new [PR 1938](https://github.com/microsoft/qlib/pull/1938). I merged the main branch based on your commit. It retains your commit information and is now merged. Thank you for your contribution to qlib. This PR will be closed.

PR #2035, opened 2025-10-16
fix yahoo collector.py

XiaotianZha:
To fix timestamp format problem
ValueError: unconverted data remains when parsing with format "%Y-%m-%d": " 09:30:00-04:00", at position 109. You might want to try:
    - passing `format` if your strings have a consistent format;
    - passing `format='ISO8601'` if your strings are all ISO8601 but not necessarily in exactly the same format;
    - passing `format='mixed'`, and the format will be inferred for each element individually. You might want to use `dayfirst` alongside this.

## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/scripts/data_collector/yahoo/collector.py
+++ b/scripts/data_collector/yahoo/collector.py
@@ -392,7 +392,8 @@ class YahooNormalize(BaseNormalize):
         columns = copy.deepcopy(YahooNormalize.COLUMNS)
         df = df.copy()
         df.set_index(date_field_name, inplace=True)
-        df.index = pd.to_datetime(df.index)
+        df.index = df.index.astype(str)
+        df.index = pd.to_datetime(df.index.str[:10], format="%Y-%m-%d")
         df.index = df.index.tz_localize(None)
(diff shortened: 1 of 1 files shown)


SunsetWolf:
Hi, @XiaotianZha 

Thanks for your attention and contribution to `qlib`,

I noticed that when I get the daily frequency data from `yahooquery`, there are some data with the time format of the last data like this `%Y-%m-%d %H:%M:%S+08:00`, which means that this data is the intraday data of the current stock, and the time displayed is the time of the last refresh. Stock data with this date format will show a similar error when normalizing: 
```
ValueError: unconverted data remains when parsing with format "%Y-%m-%d": " 09:30:00+08:00"
```
If this data is forced to be formatted as `%Y-%m-%d`, it is equivalent to treating the intraday data as the closing data, which is incorrect data for daily data.

For this problem, you can refer to the previous [PR 2015](https://github.com/microsoft/qlib/pull/2015).

PR #2062, opened 2025-11-26
feat: pytorch benchmarks support mps device

donaldkuck:
Add MPS support and centralize device selection

Add get_device() function in pytorch_utils.py supporting CUDA, MPS, and CPU
Refactor 26 PyTorch model files to use centralized device selection
Enable automatic MPS device selection on Apple Silicon devices
The device selection priority is: CUDA > MPS > CPU

### DIFF
--- a/qlib/contrib/model/pytorch_adarnn.py
+++ b/qlib/contrib/model/pytorch_adarnn.py
@@ -301,6 +301,8 @@ class ADARNN(Model):
 
         if self.use_gpu:
             torch.cuda.empty_cache()
+            if hasattr(torch.backends, "mps") and torch.backends.mps.is_available():
+                torch.mps.empty_cache()
         return best_score
 
     def predict(self, dataset: DatasetH, segment: Union[Text, slice] = "test"):
--- a/qlib/contrib/model/pytorch_add.py
+++ b/qlib/contrib/model/pytorch_add.py
@@ -415,6 +415,8 @@ class ADD(Model):
         torch.save(best_param, save_path)
         if self.use_gpu:
             torch.cuda.empty_cache()
+            if hasattr(torch.backends, "mps") and torch.backends.mps.is_available():
+                torch.mps.empty_cache()
 
     def predict(self, dataset: DatasetH, segment: Union[Text, slice] = "test"):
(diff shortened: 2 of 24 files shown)


donaldkuck:
```
FAILED model/test_general_nn.py::TestNN::test_both_dataset - RuntimeError: MPS backend out of memory (MPS allocated: 8.00 MiB, other allocations: 16.00 KiB, max allowed: 7.93 GiB). Tried to allocate 256 bytes on shared pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 to disable upper limit for memory allocations (may cause system failure).
FAILED test_contrib_model.py::TestAllFlow::test_0_initialize - RuntimeError: MPS backend out of memory (MPS allocated: 8.00 MiB, other allocations: 0 bytes, max allowed: 7.93 GiB). Tried to allocate 256 bytes on shared pool. Use PYTORCH_MPS_HIGH_WATERMARK_RATIO=0.0 to disable upper limit for memory allocations (may cause system failure).
= 2 failed, 51 passed, 1 skipped, 10 deselected, 23 warnings in 288.46s (0:04:48) =
```
I think the failure is not because of the codes.

happysl666 (approved):

SunsetWolf:
Hi, @donaldkuck , I don't think so, the `pytest` test passes in the current CI of the main branch, and in this PR `pytest` reports an error, indicating that the changes affected the `pytest` results.
The [error message](https://github.com/microsoft/qlib/actions/runs/19697499403/job/56510233610?pr=2062#step:19:1106) also points out the problem: `RuntimeError: MPS backend out of memory` due to very fragile MPS memory management.

donaldkuck:
> Hi, @donaldkuck , I don't think so, the `pytest` test passes in the current CI of the main branch, and in this PR `pytest` reports an error, indicating that the changes affected the `pytest` results. The [error message](https://github.com/microsoft/qlib/actions/runs/19697499403/job/56510233610?pr=2062#step:19:1106) also points out the problem: `RuntimeError: MPS backend out of memory` due to very fragile MPS memory management.

I think that the mac machine in CI may have only 8GB memory totally， and the memory is not enough for mps when running pytest?

PR #2077, opened 2025-12-30
refactor: implement deterministic budget allocation in SoftTopkStrategy

feedseawave:
## Description
This PR refactors the `SoftTopkStrategy` to address the non-deterministic weight allocation issue during portfolio rebalancing.

## Motivation and Context
- **Fix Portfolio Overflow Logic**: Addressed the issue where `final_stock_weight` was incorrectly handled when the number of held stocks exceeded `topk` (common when `trade_impact_limit` prevents immediate liquidation of old positions).
- **Code Cleanup**: Improved code readability by aligning comments and standardizing variable naming conventions
-  **Tests**: Added coverage for rebalancing and cold-start scenarios.

## How Has This Been Tested?

1. Pipeline test:
2. Your own tests:
test1
test2

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation


feedseawave:

feedseawave:
Hi @SunsetWolf, sorry for the message. I noticed you’ve been active on recent fixes. Just a friendly ping on this PR. I’ve checked the CI and all tests have passed. Is there any additional information or changes needed from my side to move this forward? Thanks for your time!"

feedseawave:
Hi，@you-n-g，It appears that the failure is unrelated to this PR since the same failure (ReduceLROnPlateau(verbose=...) in torch ≥2.7) in other PRs . I’ll submit a separate PR to fix the scheduler version check in pytorch_nn.py so it works across torch versions.

SunsetWolf:
> Hi，@you-n-g，It appears that the failure is unrelated to this PR since the same failure (ReduceLROnPlateau(verbose=...) in torch ≥2.7) in other PRs . I’ll submit a separate PR to fix the scheduler version check in pytorch_nn.py so it works across torch versions.

Hi, @feedseawave ,Thanks for the report, I have fixed the problem, merge main branch code is expected to fix the bug.

feedseawave:
> > Hi，@you-n-g，It appears that the failure is unrelated to this PR since the same failure (ReduceLROnPlateau(verbose=...) in torch ≥2.7) in other PRs . I’ll submit a separate PR to fix the scheduler version check in pytorch_nn.py so it works across torch versions.
> 
> Hi, @feedseawave ,Thanks for the report, I have fixed the problem, merge main branch code is expected to fix the bug.

Thanks for the update — glad the issue is resolved. I’ll close the second PR as it’s now redundant.

SunsetWolf:
Hi, @feedseawave , Thanks for the update. One thing we want to double-check before moving forward is the edge behavior under small risk_degree + partial liquidation scenarios.

In particular, there may be cases where:

- current exposure > risk_degree

- sell is capped by trade_impact_limit

- resulting budget becomes negative or insufficient for rebalance

   For example: if you start with a position (`initial_total_weight`) of 100% and a target position (`risk_degree`) of 50%, but only sell (`released_cash`) 25%, and end up with 75% left, then the `total_budget` would be -25%, a borderline case scenario.

Could you help clarify:

- Whether such cases are intentionally supported

- Or if we should explicitly guard against negative budget / over-allocation

This is mostly to make the behavior explicit for future maintainers.

feedseawave:
> Hi, @feedseawave , Thanks for the update. One thing we want to double-check before moving forward is the edge behavior under small risk_degree + partial liquidation scenarios.
> 
> In particular, there may be cases where:
> 
> * current exposure > risk_degree
> * sell is capped by trade_impact_limit
> * resulting budget becomes negative or insufficient for rebalance
>   For example: if you start with a position (`initial_total_weight`) of 100% and a target position (`risk_degree`) of 50%, but only sell (`released_cash`) 25%, and end up with 75% left, then the `total_budget` would be -25%, a borderline case scenario.
> 
> Could you help clarify:
> 
> * Whether such cases are intentionally supported
> * Or if we should explicitly guard against negative budget / over-allocation
> 
> This is mostly to make the behavior explicit for future maintainers.

Hi,SunsetWolf. Thanks for the follow-up. Indeed, the code is structured intentionally to prevent 
unintended leverages/overallocation. 

In `phase 2 : Budget allocation` ,we will first calculate `total_budget`. In the scenario you described it becomes negative so the buy phase is skipped because of the guard `if total_budget > 1e-8` This prevents initiating buys when exposure is already above target and also avoids floating‑point noise around zero.

SunsetWolf:
Hi, @feedseawave , Thanks for the contribution! We really appreciate the time and effort you put into this PR. 👍
I’ll take a final look and proceed with the merge shortly.

PR #2082, opened 2026-01-12
fix(storage): Fix FileFeatureStorage path resolution on case-sensitive file systems

### LINKED ISSUE #2053: FileFeatureStorage lowercases the instrument directory name when building feature paths
FileFeatureStorage lowercases the instrument directory name when building feature paths (instrument.lower()/field.lower().freq.bin). On case‑sensitive filesystems (e.g., Linux) where the vendor data is stored with uppercase instrument folders (data/xsto/features/ABB-U/...), the loader looks for abb-u/…, can’t find the file, and every expression resolves to an empty dataframe even though the data exists.

### To Reproduce

Store local feature data under uppercase instrument directories (e.g., data/xsto/features/ABB-U/close.day.bin).
Initialize qlib with that provider URI.
Request any feature for that instrument, e.g. D.features(["ABB-U"], ["$close"], "2000-01-01", "2025-12-31"). The result is an empty dataframe and the feature storage backend reports the file is missing.

### Expected Behavior
FileFeatureStorage should respect the actual on-disk casing instrument list (or probe common case variants) so that features load correctly regardless of filesystem case sensitivity.

### Screenshot
N/A – the console logs show statements such as [FileFeatureStorage.__getitem__] uri=…/features/abb-u/close.day.bin missing for request=… even though …/features/ABB-U/close.day.bin exists.

### Environment
Version: 0.9.7
Python version: 3.12 (virtual env)
OS: Linux (Ubuntu on AMD cloud server, case-sensitive filesystem)
Commit number: N/A (using pip wheel)
Additional Notes
We patched FileFeatureStorage locally to keep multiple filename candidates (instrument.lower(), original case, uppercase) and select the first existing path before reading. After this change ABB-U and other uppercase symbols load correctly. Happy to submit a PR if this approach sounds acceptable.

im-anishraj:
## Summary
Fixes #2053: FileFeatureStorage incorrectly lowercases instrument directory names, causing failures on case-sensitive file systems.

## What I changed
- Updated `qlib/data/storage/file_storage.py` to resolve instrument directories using the correct filesystem case when available.
- Added a fallback to lowercase paths for backward compatibility.
- Ensured new file creation preserves the original instrument case.
- Added `tests/storage_tests/test_issue_2053.py` to validate behavior on case-sensitive systems.

## How I tested
- Added an isolated test (`test_issue_2053.py`) that reproduces the issue by creating an uppercase instrument directory.
- Verified that `FileFeatureStorage` resolves paths correctly after the fix.
- Note: Full test suite was not executed locally due to heavy optional dependencies; CI should validate integration.

## Before / After
**Before**
`FileFeatureStorage(instrument="AAPL").uri` → `.../aapl/close.day.bin` (fails if directory is `AAPL`)

**After**
`FileFeatureStorage(instrument="AAPL").uri` → `.../AAPL/close.day.bin`

Fixes #2053

### DIFF
--- a/qlib/data/storage/file_storage.py
+++ b/qlib/data/storage/file_storage.py
@@ -303,8 +303,13 @@ class FileFeatureStorage(FileStorageMixin, FeatureStorage):
         name_orig = f"{self.instrument}/{self.field.lower()}.{self.freq.lower()}.bin"
         if (base_uri / name_orig).exists():
             return name_orig
+
+        # Candidate 2: Uppercase (Fix for lowercase input finding uppercase folder on case-sensitive OS)
+        name_upper = f"{self.instrument.upper()}/{self.field.lower()}.{self.freq.lower()}.bin"
+        if (base_uri / name_upper).exists():
+            return name_upper
             
-        # Candidate 2: Lowercase (Backward Compatibility)
+        # Candidate 3: Lowercase (Backward Compatibility)
         name_lower = f"{self.instrument.lower()}/{self.field.lower()}.{self.freq.lower()}.bin"
         if (base_uri / name_lower).exists():
             return name_lower


im-anishraj:

SunsetWolf:
Hi @anish-devgit , Thank you so much for taking the time to contribute and for providing a thoughtful solution! We really appreciate your effort.

In this case, `Qlib` intentionally normalizes instrument names to lowercase to ensure consistent behavior across different platforms, especially Windows. Allowing multiple case variants in `FileFeatureStorage` would break that consistency and could introduce subtle cross-platform issues.

Thanks again for the contribution and for offering to submit a PR — we really value the discussion and grateful for your understanding.

im-anishraj:
Thanks a lot for the detailed explanation and for reviewing my PR 🙏  
That makes sense — I understand the design decision around normalizing to lowercase for cross-platform consistency.  

Appreciate the guidance and the kind feedback. Happy to contribute again in the future!

PR #2083, opened 2026-01-12
fix(online): add interval check to RollingStrategy to prevent premature task generation (#2045)

### LINKED ISSUE #2045: [Question] Why not check if the time interval exceeds the step size
https://github.com/microsoft/qlib/blob/bb7ab1cf143b93a1656d13cfdbd40d1e7c15edd3/qlib/workflow/online/strategy.py#L181C9-L188C75

Why not check if the time interval exceeds the step size? And if it doesn't, shouldn't we return an empty list?

interval = self.ta.cal_interval(calendar_latest, max_test[0])
if interval < self.rg.step:
    return []

im-anishraj:
Summary
Fixes #2045: RollingStrategy should not prepare tasks if interval < rolling_step

What I changed
- Modified `qlib/workflow/online/strategy.py`: Added a check in `prepare_tasks` to return specific empty list if the calculated time interval is less than `self.rg.step`.

How I tested
Local: Added `tests/test_issue_2045.py` which mocks `RollingGen`, `TimeAdjuster`, and `OnlineToolR`.
- Verified that when `interval=5` and `step=10`, `prepare_tasks` returns `[]`.
- Verified that when `interval=15` and `step=10`, `prepare_tasks` returns generated tasks.

Fixes #2045

### DIFF
--- a/qlib/data/storage/file_storage.py
+++ b/qlib/data/storage/file_storage.py
@@ -286,7 +286,29 @@ class FileFeatureStorage(FileStorageMixin, FeatureStorage):
     def __init__(self, instrument: str, field: str, freq: str, provider_uri: dict = None, **kwargs):
         super(FileFeatureStorage, self).__init__(instrument, field, freq, **kwargs)
         self._provider_uri = None if provider_uri is None else C.DataPathManager.format_provider_uri(provider_uri)
-        self.file_name = f"{instrument.lower()}/{field.lower()}.{freq.lower()}.bin"
+        
+        base_uri = self.dpm.get_data_uri(self.freq) # Assuming dpm is available and get_data_uri provides the base path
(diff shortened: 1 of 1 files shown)


SunsetWolf:
Hi, @anish-devgit , Thanks for the contribution! The concern makes sense, but we intentionally avoid checking `interval < step` here. This logic is handled inside `RollingGen.gen_following_tasks`, and duplicating it at the strategy level would be redundant.

For these reasons, we’ll close this PR for now, , but we really appreciate the thoughtful contribution.

im-anishraj:
Got it, thanks for the clear explanation!
Makes sense to keep this logic centralized in RollingGen.gen_following_tasks.
Appreciate the review 🙌

PR #2086, opened 2026-01-14
fix(security): prevent path traversal in user data deletion

tkiran:
Problem: UserManager.remove_user() built user_path = self.data_path / user_id and then called shutil.rmtree(user_path). Because user_id is user-controlled, an attacker could supply traversal/absolute paths (e.g., ../.., C:\Windows, \\server\share) to delete directories outside the intended user data directory.

Impact: Arbitrary recursive directory deletion (critical), potentially wiping system files or other application data.

Fix: Added strict user_id validation and a safe path resolver that:

- rejects traversal, absolute/UNC/drive paths, separators, and null bytes
- resolves paths and enforces that the target directory is contained within self.data_path (prevents traversal and symlink escapes)
- uses this safe path in remove_user (and consistently across related user file operations)

### DIFF
--- a/qlib/contrib/online/manager.py
+++ b/qlib/contrib/online/manager.py
@@ -7,6 +7,7 @@
 import pathlib
 import pandas as pd
 import shutil
+import os
 from ruamel.yaml import YAML
 from ...backtest.account import Account
 from .user import User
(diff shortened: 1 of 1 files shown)


SunsetWolf:
Hi, @tkiran , Thanks for the contribution! Before we can proceed, could you please review and sign the Contributor License Agreement (CLA)? Once that’s done, we’ll be happy to move forward with the PR. Thanks!

PR #2135, opened 2026-02-15
Fix FileFeatureStorage path resolution on case-sensitive filesystems

### LINKED ISSUE #2053: FileFeatureStorage lowercases the instrument directory name when building feature paths

pdwi2020:
## Summary
- keep the existing lowercase path as default for feature storage
- add case-aware fallback that resolves existing instrument directories/files by case-insensitive match
- add a regression test for uppercase instrument directory (`ABB-U`) with `close.day.bin`

## Why
This fixes loading failures on case-sensitive filesystems when data is stored under uppercase instrument folders, as reported in #2053.

Fixes #2053

## Testing
- `python3 -m py_compile qlib/data/storage/file_storage.py tests/storage_tests/test_storage.py`
- Full pytest target requires compiled qlib C extensions in this environment.

### DIFF
--- a/qlib/data/storage/file_storage.py
+++ b/qlib/data/storage/file_storage.py
@@ -3,7 +3,7 @@
 
 import struct
 from pathlib import Path
-from typing import Iterable, Union, Dict, Mapping, Tuple, List
+from typing import Iterable, Union, Dict, Mapping, Tuple, List, Optional
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


pdwi2020:

PR #2161, opened 2026-03-21
fix(yahoo): use adjclose for $change calculation

### LINKED ISSUE #1563: $CHANGE is calculated based on close instead of adj_close in the YahooCollector
## 🐛 Bug Description

$CHANGE is calculated based on close instead of adj_close in the YahooCollector

## To Reproduce

1. Download the data from [Yahoo Finance](https://finance.yahoo.com/quote/301303.SZ/history?p=301303.SZ). Using 301303.SZ as an example only because the dataset is small.
2. Notice the following data in the downloaded csv (removed unnecessary columns for demonstration purpose)
`Date,Close,Adj Close
2023-06-05,23.120001,22.893938
2023-06-06,22.500000,22.280001
2023-06-07,22.240000,22.240000`
3. Normalize the downloaded csv file using `data_collector.yahoo.collector.YahooNormalizeCN1d`.
4. Check the normalized data. The `change` on 2023-06-06 is `-0.026816608996539815` which is correct and the `change` on 2023-06-07 is `-0.011555555555555652` which is wrong. 

## Expected Behavior

The `change` on 2023-06-07 should be `22.240000/22.280001 - 1. = -0.0017953769391662044`.

## Environment

Linux x86_64
Python version: 3.8.16 
Qlib version: 0.9.1

Bortlesboat:
## Summary

Fixes #1563. `YahooNormalize.calc_change()` previously used raw close prices, so splits and dividends could create artificial return spikes that were then interpreted as limit-up or limit-down events during backtests.

- Use `adjclose` when Yahoo provides it.
- Retain `close` as a fallback for sources without adjusted prices.
- Add focused regressions for split-adjusted and fallback behavior.

## Verification

- `python -m pytest tests/test_yahoo_collector.py -q` — 2 passed
- Ruff and formatting checks pass for the new test.
- Diff whitespace check passes.

### DIFF
--- a/scripts/data_collector/yahoo/collector.py
+++ b/scripts/data_collector/yahoo/collector.py
@@ -371,7 +371,9 @@ class YahooNormalize(BaseNormalize):
     @staticmethod
     def calc_change(df: pd.DataFrame, last_close: float) -> pd.Series:
         df = df.copy()
-        _tmp_series = df["close"].ffill()
+        # Use adjusted close to correctly account for splits/dividends
+        _close_col = "adjclose" if "adjclose" in df.columns else "close"
+        _tmp_series = df[_close_col].ffill()
         _tmp_shift_series = _tmp_series.shift(1)
         if last_close is not None:
             _tmp_shift_series.iloc[0] = float(last_close)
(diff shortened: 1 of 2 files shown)


Bortlesboat:
Rebased onto current main and added focused regression coverage for both a split discontinuity and the raw-close fallback. Both tests pass locally; the new test also passes Ruff and formatting checks.

PR #2164, opened 2026-03-22
fix(model): use `in` instead of `==` for metric comparison in HIST and IGMTF

### LINKED ISSUE #2163: Bug: metric_fn in HIST and IGMTF always raises ValueError for default metric
### Description

`metric_fn()` in `pytorch_hist.py` (line 173) and `pytorch_igmtf.py` (line 166) uses `==` to compare `self.metric` against the tuple `("", "loss")`:

```python
if self.metric == ("", "loss"):
    return -self.loss_fn(pred[mask], label[mask])
```

Since `self.metric` is always a string (default is `""`), this comparison is never true. The function falls through to `raise ValueError("unknown metric")`.

All other models correctly use `in`:

```python
# pytorch_alstm.py, pytorch_gru.py, pytorch_lstm.py, etc.
if self.metric in ("", "loss"):
    return -self.loss_fn(pred[mask], label[mask])
```

### Impact

Any user training HIST or IGMTF with the default metric (or `metric="loss"`) will get a crash:
```
ValueError: unknown metric ``
```

### Fix

```diff
- if self.metric == ("", "loss"):
+ if self.metric in ("", "loss"):
```

I will submit a fix shortly.

warren618:
## Summary

Fixes #2163

`metric_fn()` in `pytorch_hist.py` and `pytorch_igmtf.py` compares `self.metric` against the tuple `("", "loss")` using `==`. Since `self.metric` is always a string, this condition is never true, causing `ValueError("unknown metric")` on the default metric path.

All other models (ALSTM, GRU, LSTM, ALSTM_TS) correctly use `in ("", "loss")`.

## Changes

- `qlib/contrib/model/pytorch_hist.py`: `==` → `in` (line 173)
- `qlib/contrib/model/pytorch_igmtf.py`: `==` → `in` (line 166)

## Before / After

```python
# Before (broken): string == tuple is always False
if self.metric == ("", "loss"):

# After (correct): matches when metric is "" or "loss"
if self.metric in ("", "loss"):
```

### DIFF
--- a/qlib/contrib/model/pytorch_hist.py
+++ b/qlib/contrib/model/pytorch_hist.py
@@ -170,7 +170,7 @@ class HIST(Model):
             vy = y - torch.mean(y)
             return torch.sum(vx * vy) / (torch.sqrt(torch.sum(vx**2)) * torch.sqrt(torch.sum(vy**2)))
 
-        if self.metric == ("", "loss"):
+        if self.metric in ("", "loss"):
             return -self.loss_fn(pred[mask], label[mask])
 
         raise ValueError("unknown metric `%s`" % self.metric)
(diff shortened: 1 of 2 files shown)


warren618:

PR #2168, opened 2026-03-22
fix(model): replace mutable default evals_result=dict() with None across 27 models

### LINKED ISSUE #2167: Bug: mutable default argument evals_result=dict() in 27 model fit() methods
### Description

27 models in `qlib/contrib/model/` use `evals_result=dict()` as a default argument in their `fit()` method. This is a [well-known Python antipattern](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments): the dict is created once at function definition time and shared across all calls, causing potential state leakage between separate `fit()` invocations.

The correct pattern (already used in `gbdt.py` and `highfreq_gdbt_model.py`) is:

```python
def fit(self, dataset, evals_result=None):
    if evals_result is None:
        evals_result = {}
```

### Affected files

All 27 files in `qlib/contrib/model/` except `gbdt.py` and `highfreq_gdbt_model.py`.

I will submit a fix shortly.

warren618:
## Summary

Fixes #2167

27 models use `evals_result=dict()` as a mutable default argument in `fit()`. Changed to `evals_result=None` with explicit guard, matching the pattern in `gbdt.py`.

## Changes

For each of the 27 affected files:
```diff
- def fit(self, dataset, evals_result=dict()):
+ def fit(self, dataset, evals_result=None):
+     if evals_result is None:
+         evals_result = {}
```

27 files changed, 81 insertions, 27 deletions.

### DIFF
--- a/qlib/contrib/model/catboost_model.py
+++ b/qlib/contrib/model/catboost_model.py
@@ -31,10 +31,12 @@ class CatBoostModel(Model, FeatureInt):
         num_boost_round=1000,
         early_stopping_rounds=50,
         verbose_eval=20,
-        evals_result=dict(),
+        evals_result=None,
         reweighter=None,
         **kwargs,
     ):
+        if evals_result is None:
+            evals_result = {}
         df_train, df_valid = dataset.prepare(
             ["train", "valid"],
             col_set=["feature", "label"],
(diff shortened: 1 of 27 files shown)


warren618:

PR #2170, opened 2026-03-22
fix(model): add pretrain toggle to HIST model

### LINKED ISSUE #2074: HIST model missing explicit pretrain toggle in initialization
In contrib\model\pytorch_hist.py
Description: Currently, the HIST model implementation in qlib lacks an explicit parameter to control the pre-training loading logic. This makes it difficult for users to skip the pre-training weight-loading phase even when a model path is specified or when they wish to train the architecture from scratch.

Proposed Changes:

__init__ method: Added pretrain (defaulting to True for backward compatibility) to the parameters.

Logic Update: Wrapped the weight-loading and state_dict update logic inside if self.pretrain:.
```
# In __init__
def __init__(self, ..., pretrain=True, **kwargs):
    ...
    self.pretrain = pretrain

# In weight loading logic
if self.pretrain:
    if self.model_path is not None:
        self.logger.info("Loading pretrained model...")
        pretrained_model.load_state_dict(torch.load(self.model_path))
    
    # ... (state_dict update logic)
    self.logger.info("Loading pretrained model Done...")
```

warren618:
## Summary

Fixes #2074

Added a `pretrain` parameter (default `True` for backward compatibility) to HIST model. When `pretrain=False`, the model trains from scratch without loading pretrained base_model weights.

## Changes

- `qlib/contrib/model/pytorch_hist.py`:
  - Add `pretrain=True` parameter to `__init__`
  - Wrap weight-loading logic in `if self.pretrain:`

## Usage

```python
# Default: loads pretrained weights (backward compatible)
model = HISTModel(model_path="path/to/weights")

# Train from scratch
model = HISTModel(pretrain=False)
```

### DIFF
--- a/qlib/contrib/model/pytorch_hist.py
+++ b/qlib/contrib/model/pytorch_hist.py
@@ -59,6 +59,7 @@ class HIST(Model):
         optimizer="adam",
         GPU=0,
         seed=None,
+        pretrain=True,
         **kwargs,
     ):
         # Set logger.
@@ -82,6 +83,7 @@ class HIST(Model):
         self.stock_index = stock_index
         self.device = torch.device("cuda:%d" % (GPU) if torch.cuda.is_available() and GPU >= 0 else "cpu")
         self.seed = seed
+        self.pretrain = pretrain
 
         self.logger.info(
             "HIST parameters setting:"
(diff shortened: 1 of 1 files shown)


warren618:

PR #2171, opened 2026-03-27
fix: TanhProcess using wrong MultiIndex column level

### LINKED ISSUE #1687: class TanhProcess(Processor) error
## 🐛 Bug Description

data.columns.get_level_values(1).str.contains("LABEL") is weird.
data.columns.get_level_values(1) will never contain LABEL, but data.columns.get_level_values(0) will.
if you want filter label, why do not use fields_group params?

class TanhProcess(Processor):
    """Use tanh to process noise data"""

    def __call__(self, df):
        def tanh_denoise(data):
            mask = data.columns.get_level_values(1).str.contains("LABEL")
            col = df.columns[~mask]
            data[col] = data[col] - 1
            data[col] = np.tanh(data[col])

            return data

        return tanh_denoise(df)

## To Reproduce

1.
1.
1.

## Environment

and paste them here directly.

 - Qlib version:
 - Python version:

EdenRochmanSharabi:
## Summary

Fixes #1687

`TanhProcess.__call__()` uses `get_level_values(1)` to identify LABEL columns, but LABEL group names appear at level 0 of the MultiIndex, not level 1. This causes the tanh denoising to be incorrectly applied to label columns.

## Changes

- `qlib/data/dataset/processor.py`: Changed `get_level_values(1)` to `get_level_values(0)`, and fixed variable reference from `df` to `data` for the column mask
- `tests/test_tanh_process.py`: Test verifying LABEL columns are untouched while FEATURE columns are tanh-transformed

### DIFF
--- a/qlib/data/dataset/processor.py
+++ b/qlib/data/dataset/processor.py
@@ -148,8 +148,8 @@ class TanhProcess(Processor):
 
     def __call__(self, df):
         def tanh_denoise(data):
-            mask = data.columns.get_level_values(1).str.contains("LABEL")
-            col = df.columns[~mask]
+            mask = data.columns.get_level_values(0).str.contains("LABEL")
+            col = data.columns[~mask]
             data[col] = data[col] - 1
             data[col] = np.tanh(data[col])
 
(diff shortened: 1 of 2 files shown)


EdenRochmanSharabi:

PR #2172, opened 2026-03-27
fix: TCN model predict crash on single-sample final batch

### LINKED ISSUE #1752: qlib/qlib/contrib/model /pytorch_tcn_ts.py error
## To Reproduce

1.  set batch size of model as 2000
1.  set len of infer dataset  as 10001
1.  error occur

reason: infer dataset will be split as 5*2000+1, 2000 will be shaped as np.array vector, and 1 will be shaped as np.array scalar.
Calling np.concatenate between vector and scalar will raise error.

## Screenshot

image

## Environment

and paste them here directly.

 - Qlib version:
 - Python version:

EdenRochmanSharabi:
## Summary

Fixes #1752

When the inference dataset size is not evenly divisible by batch size, the last batch has 1 sample. `squeeze()` produces a 0-d tensor, and after `.numpy()`, `np.concatenate` fails because it cannot concatenate 0-d arrays with 1-d arrays.

## Changes

- `qlib/contrib/model/pytorch_tcn_ts.py`: Wrapped prediction output with `np.atleast_1d()` before appending to the results list
- `tests/test_tcn_single_sample_batch.py`: Tests verifying concatenation works with mixed array dimensions

### DIFF
--- a/qlib/contrib/model/pytorch_tcn_ts.py
+++ b/qlib/contrib/model/pytorch_tcn_ts.py
@@ -279,7 +279,7 @@ class TCN(Model):
             with torch.no_grad():
                 pred = self.TCN_model(feature.float()).detach().cpu().numpy()
 
-            preds.append(pred)
+            preds.append(np.atleast_1d(pred))
 
         return pd.Series(np.concatenate(preds), index=dl_test.get_index())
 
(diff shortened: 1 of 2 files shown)


EdenRochmanSharabi:

PR #2183, opened 2026-04-04
fix(model): replace mutable default evals_result=dict() with None in 28 fit() methods

### LINKED ISSUE #2167: Bug: mutable default argument evals_result=dict() in 27 model fit() methods

abhiojha8:
Fixes #2167

## Summary

- Replace `evals_result=dict()` with `evals_result=None` + explicit guard in **28 model `fit()` methods** across `qlib/contrib/model/` and `examples/benchmarks/TRA/src/model.py`
- Add regression test using AST parsing to prevent reintroduction

## Problem

Using `evals_result=dict()` as a default argument is a [well-known Python antipattern](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments): the dict is created once at function definition time and shared across all calls, causing state leakage between separate `fit()` invocations.

## Fix

Applied the same pattern already used in `gbdt.py` and `highfreq_gdbt_model.py`:

```python
# Before
def fit(self, dataset, evals_result=dict()):

# After
def fit(self, dataset, evals_result=None):
    if evals_result is None:
        evals_result = {}
```

## Verification

- All 28 affected files fixed consistently
- `tests/test_contrib_model.py` passes (model initialization)
- `tests/misc/` test suite passes
- New regression test (`tests/test_mutable_default_evals_result.py`) passes
- black and flake8 checks pass

### DIFF
--- a/examples/benchmarks/TRA/src/model.py
+++ b/examples/benchmarks/TRA/src/model.py
@@ -200,7 +200,9 @@ class TRAModel(Model):
 
         return metrics, preds
 
-    def fit(self, dataset, evals_result=dict()):
+    def fit(self, dataset, evals_result=None):
+        if evals_result is None:
+            evals_result = {}
         train_set, valid_set, test_set = dataset.prepare(["train", "valid", "test"])
 
         best_score = -1
--- a/qlib/contrib/model/catboost_model.py
+++ b/qlib/contrib/model/catboost_model.py
@@ -31,10 +31,12 @@ class CatBoostModel(Model, FeatureInt):
         num_boost_round=1000,
         early_stopping_rounds=50,
         verbose_eval=20,
-        evals_result=dict(),
+        evals_result=None,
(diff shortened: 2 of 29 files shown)


abhiojha8:

PR #2189, opened 2026-04-10
feat(evaluate): add Sortino, Calmar, and max drawdown duration to risk_analysis

JKDasondee:
## Summary

Extends `risk_analysis()` in `qlib/contrib/evaluate.py` with three standard risk metrics that every quant platform should expose:

| Metric | Formula | Edge case |
|--------|---------|-----------|
| `sortino_ratio` | `annualized_return / downside_std * sqrt(N)` | `inf` when no negative returns |
| `calmar_ratio` | `annualized_return / abs(max_drawdown)` | `inf` when no drawdown |
| `max_drawdown_duration` | longest period below cumulative peak (in observations) | `0` when no drawdown |

These are the three most commonly used risk-adjusted performance metrics after Sharpe/IR and max drawdown, which `risk_analysis` already computes.

## Implementation

19 lines added, no new dependencies. All three metrics:
- Respect the existing `mode` parameter (`"sum"` vs `"product"` accumulation)
- Use the same annualization scaler `N`
- Are appended to the existing output `pd.Series`, so existing code that reads `risk_analysis()` output is fully backwards compatible

Downside deviation uses `r.clip(upper=0).std(ddof=1)` (standard Sortino convention — only negative returns contribute to risk).

Max drawdown duration counts contiguous periods where the cumulative curve is below its running maximum, using `cummax()` group labeling.

## Verification

```python
r = pd.Series(np.random.randn(252) * 0.01)
res = risk_analysis(r, freq='day')

# Manual Sortino matches
ds = r.clip(upper=0).std(ddof=1)
assert abs(res.loc['sortino_ratio', 'risk'] - r.mean() / ds * np.sqrt(238)) < 1e-10

# Manual Calmar matches
ann_ret = r.mean() * 238
max_dd = (r.cumsum() - r.cumsum().cummax()).min()
assert abs(res.loc['calmar_ratio', 'risk'] - ann_ret / abs(max_dd)) < 1e-10

# Edge cases
r_pos = pd.Series(np.abs(np.random.randn(100)) * 0.01)
assert risk_analysis(r_pos, freq='day').loc['calmar_ratio', 'risk'] == np.inf

# Product mode
res_prod = risk_analysis(r, freq='day', mode='product')
assert 'sortino_ratio' in res_prod.index
```

Existing `tests/ops/` and `tests/misc/` suites pass (18 tests, 0 failures).

### DIFF
--- a/qlib/contrib/evaluate.py
+++ b/qlib/contrib/evaluate.py
@@ -82,12 +82,31 @@ def risk_analysis(r, N: int = None, freq: str = "day", mode: Literal["sum", "pro
         raise ValueError(f"risk_analysis accumulation mode {mode} is not supported. Expected `sum` or `product`.")
 
     information_ratio = mean / std * np.sqrt(N)
+
+    downside = r.clip(upper=0)
+    downside_std = downside.std(ddof=1)
+    sortino_ratio = mean / downside_std * np.sqrt(N) if downside_std > 0 else np.inf
+    calmar_ratio = annualized_return / abs(max_drawdown) if max_drawdown != 0 else np.inf
+
+    if mode == "sum":
+        cum = r.cumsum()
+    else:
+        cum = (1 + r).cumprod()
+    peak = cum.cummax()
+    in_drawdown = cum < peak
+    dd_groups = (~in_drawdown).cumsum()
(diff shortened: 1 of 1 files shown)


JKDasondee:

PR #2202, opened 2026-05-05
fix: HDFStore Resource Leak in DiskDatasetCache

Sagar-py:
## Description
Replace manual `pd.HDFStore` open/close with a context manager in `DiskDatasetCache.update()` (`qlib/data/cache.py`)

## Motivation and Context
The previous pattern relied on an explicit `store.close()` call that was never reached if any intermediate operation raised an exception. This leaked file descriptors and could leave the cache file in a partially-written corrupted state - making it all subsequent reads from that cache fail until the file was manually removed.

## How Has This Been Tested?
- [X] Pass the test by running: `pytest qlib/tests/test_all_pipeline.py` under upper directory of `qlib`.

## Types of changes
- [X] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/data/cache.py
+++ b/qlib/data/cache.py
@@ -1031,21 +1031,20 @@ class DiskDatasetCache(DatasetCache):
                 else:
                     return 0  # No data to update cache
 
-                store = pd.HDFStore(cp_cache_uri)
                 # FIXME:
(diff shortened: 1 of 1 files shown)


Sagar-py:

PR #2212, opened 2026-05-12
fix(config): keep QlibConfig.registered safe in joblib worker subprocesses

### LINKED ISSUE #2038: Backtest/joblib worker crash: AttributeError: No such registered in self._config
## 🐛 Bug Description
During backtests that spawn joblib workers, fresh worker processes import Qlib and call config.register_from_C(g_config). That function checks C.registered before copying g_config. Because worker C._config lacks the key on first access, Config.getattr raises:
AttributeError: No such registered in self._config

Proposed fix
In qlib/config.py register_from_C:
Replace:
if C.registered and skip_register:
With:
if getattr(C, "registered", False) and skip_register:
After merging g_config into the worker:
self._config["registered"] = True

## To Reproduce
qlib.backtest.report.PortfolioMetrics._cal_benchmark → qlib.utils.resam.get_higher_eq_freq_feature → D.features → DatasetD.dataset → dataset_processor → joblib.Parallel → qlib.data.data.inst_calculator → qlib.config.C.register_from_C
qlib/config.py: register_from_C:
if C.registered and skip_register: # raises AttributeError (key missing in worker)

## Expected Behavior
Worker config should safely sync from the main process without assuming the key exists.

## Environment
OS: Windows 11
Python: 3.10
Qlib: 0.9.7
Joblib: version bundled with Qlib deps
Backend: default process-based Parallel

## Additional Notes

This avoids needing user-side hacks or thread-only backends. Defaulting the flag to False on first access is consistent with intent and prevents the worker crash.

genisis0x:
## Summary
- Fixes #2038. `QlibConfig.registered` is implemented as a `@property` that returns `self._registered`, but because the overridden `__setattr__` routes attribute writes through `self.__dict__["_config"]`, the `_registered` flag actually lives inside the `_config` dict. If `_config` is rebuilt from `_default_config` without the `_registered` key (e.g. inside a freshly-spawned joblib worker that re-imports `qlib`, or transiently inside `Config.reset()` before `QlibConfig.__init__` re-sets the flag), the property's inner `self._registered` access raises `AttributeError`, Python falls back to `Config.__getattr__` for the original attribute name `registered`, and the worker dies with:

  ```
  AttributeError: No such ``registered`` in self._config
  ```

  This terminates backtests along `PortfolioMetrics._cal_benchmark` -> `get_higher_eq_freq_feature` -> `D.features` -> `ParallelExt` -> `inst_calculator` -> `register_from_C`.

- Make the property defensive: read `_registered` out of `self.__dict__["_config"]` via a safe `.get()` chain so a missing key returns `False` instead of triggering the property -> `__getattr__` cascade.
- Harden `register_from_C` with `getattr(C, "registered", False)` so the second top-level caller (`qlib/config.py:115`) is double-protected against the same failure mode.

## Validation
- `uv run pytest tests/misc/test_config_registered.py` -- 4 passed.
- Pre-fix sanity check: stashing the patch and running the new test suite reproduces the reporter's exact error message (`AttributeError: No such ``registered`` in self._config`) on `test_registered_property_returns_false_when_key_missing` and `test_registered_property_returns_false_when_config_missing`; reapplying the patch turns both green.
- `uv run black qlib/config.py tests/misc/test_config_registered.py -l 120 --check --diff` -- clean.
- `uv run flake8 --ignore=E501,F541,E266,E402,W503,E731,E203 qlib/config.py` -- clean.

## Tests added
- `test_registered_property_returns_false_when_key_missing` -- worker scenario where `_config` lacks the `_registered` key.
- `test_registered_property_returns_false_when_config_missing` -- partially-constructed instance (e.g. mid-unpickle) where even `_config` is absent.
- `test_register_from_C_does_not_raise_when_registered_missing` -- exact guarded boolean from the original bug report does not raise on the global `C` when the key is removed.
- `test_registered_property_after_normal_construction` -- happy path still reflects the assigned value.

Fixes #2038

### DIFF
--- a/qlib/config.py
+++ b/qlib/config.py
@@ -112,7 +112,16 @@ class Config:
     def register_from_C(config, skip_register=True):
         from .utils import set_log_with_config  # pylint: disable=C0415
 
-        if C.registered and skip_register:
+        # ``C.registered`` resolves through ``QlibConfig.registered``, which
+        # reads ``self._registered`` -- a key stored inside ``self._config`` via
+        # the overridden ``__setattr__``.  In freshly-spawned joblib worker
+        # processes that key may be absent (e.g. when ``Config.reset()`` rebuilt
+        # ``_config`` from ``_default_config`` between ``QlibConfig.__init__``
+        # and this call), and ``Config.__getattr__`` would then raise
+        # ``AttributeError: No such ``registered`` in self._config`` and crash
+        # the backtest worker (issue #2038).  Falling back to ``False`` lets
+        # the worker re-register safely instead of dying.
+        if getattr(C, "registered", False) and skip_register:
             return
 
         C.set_conf_from_C(config)
@@ -520,7 +529,14 @@ class QlibConfig(Config):
 
     @property
     def registered(self):
-        return self._registered
+        # ``_registered`` is stored inside ``self._config`` because the
+        # overridden ``__setattr__`` routes attribute writes there.  After a
+        # ``Config.reset()`` (or in joblib worker subprocesses where ``_config``
+        # is recreated from ``_default_config`` before ``QlibConfig.__init__``
+        # finishes), the key may be missing.  Treat missing as ``False`` so
+        # callers see a consistent "not yet registered" signal instead of
+        # ``AttributeError`` (issue #2038).
+        return self.__dict__.get("_config", {}).get("_registered", False)
 
 
 # global config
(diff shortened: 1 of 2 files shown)


genisis0x:
Read through the CLA — all good.

genisis0x:
Hi @SunsetWolf — this PR has been open ~7 days with no review yet. Fixes a `QlibConfig.registered` issue in joblib workers — when the worker process re-initialises qlib, the `registered` flag gets carried over from the parent and prevents fresh registration. Would you have a moment to take a look? Thanks.

PR #2213, opened 2026-05-12
fix(unpickler): allow Alpha158/Alpha360 handlers and the standard dataset chain

### LINKED ISSUE #2130: UnpicklingError: Forbidden class: qlib.contrib.data.handler.Alpha158
### Env
pyqlib 0.9.8.dev26

### Description

The error
```
UnpicklingError: Forbidden class: qlib.contrib.data.handler.Alpha158. Only whitelisted classes are allowed for security reasons. This is to prevent arbitrary code execution through pickle deserialization.
```
only happens in rolling train, i.e. `Rolling._train_rolling_tasks`,
not happen in unrolling train.

When downgrapde pyqlib to 0.9.7, the error disappears.

genisis0x:
## Summary
- Fixes #2130. The RestrictedUnpickler safelist landed by the recent security hardening (#2099 / #2076 / #2153) only covered the abstract `DataHandler` / `DataHandlerLP` classes plus `StaticDataLoader`. Any rolling workflow that pickles a real Dataset (the default for `Rolling._train_rolling_tasks`) walks into one of the contrib stock handlers and now crashes on reload:

  ```
  UnpicklingError: Forbidden class: qlib.contrib.data.handler.Alpha158.
  Only whitelisted classes are allowed for security reasons. ...
  ```

  Unrolling workflows happened to use a path that did not go through the restricted loader, which is why downgrading from `0.9.8.dev26` to `0.9.7` hid the issue.

- Extend `SAFE_PICKLE_CLASSES` with the qlib-internal classes that sit on the standard recorder pickle graph:
  - The four shipped contrib handlers: `Alpha158`, `Alpha158vwap`, `Alpha360`, `Alpha360vwap`.
  - The dataset wrappers (`Dataset`, `DatasetH`, `TSDatasetH`) and the additional concrete loaders (`DataLoader`, `DLWParser`, `QlibDataLoader`, `NestedDataLoader`, `DataLoaderDH`).
  - Every concrete `Processor` defined in `qlib.data.dataset.processor` -- they appear in every realistic `learn_processors` / `infer_processors` chain.

- These are all classes already shipped inside qlib itself, so adding them does not weaken the threat model the safelist was designed against (arbitrary code execution through externally-supplied pickle payloads).

## Validation
- `uv run pytest tests/misc/test_pickle_safelist.py tests/misc/test_config_registered.py` -- 13 passed (25 subtests).
- `uv run black qlib/utils/pickle_utils.py tests/misc/test_pickle_safelist.py -l 120 --check --diff` -- clean.
- `uv run flake8 --ignore=E501,F541,E266,E402,W503,E731,E203 qlib/utils/pickle_utils.py` -- clean.
- End-to-end check: `RestrictedUnpickler.find_class("qlib.contrib.data.handler", "Alpha158")` now returns the real `Alpha158` class object instead of raising, and a payload containing an unrelated unknown class is still rejected.

Fixes #2130

### DIFF
--- a/qlib/utils/pickle_utils.py
+++ b/qlib/utils/pickle_utils.py
@@ -46,9 +46,45 @@ SAFE_PICKLE_CLASSES: Set[Tuple[str, str]] = {
     ("pathlib", "Path"),
     ("pathlib", "PosixPath"),
     ("pathlib", "WindowsPath"),
+    ("qlib.data.dataset.handler", "DataHandlerABC"),
     ("qlib.data.dataset.handler", "DataHandler"),
     ("qlib.data.dataset.handler", "DataHandlerLP"),
+    ("qlib.data.dataset.loader", "DataLoader"),
+    ("qlib.data.dataset.loader", "DLWParser"),
+    ("qlib.data.dataset.loader", "QlibDataLoader"),
     ("qlib.data.dataset.loader", "StaticDataLoader"),
+    ("qlib.data.dataset.loader", "NestedDataLoader"),
+    ("qlib.data.dataset.loader", "DataLoaderDH"),
+    # Dataset hierarchy - needed when a recorder/rolling workflow pickles a
+    # full dataset and the unpickler walks the wrapped handler/loader graph.
+    ("qlib.data.dataset", "Dataset"),
+    ("qlib.data.dataset", "DatasetH"),
+    ("qlib.data.dataset", "TSDatasetH"),
+    # Stock-data handlers shipped in qlib.contrib. Without these the
+    # ``Rolling._train_rolling_tasks`` -> recorder load path fails with
+    # ``Forbidden class: qlib.contrib.data.handler.Alpha158`` (issue #2130).
+    ("qlib.contrib.data.handler", "Alpha158"),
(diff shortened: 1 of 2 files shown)


genisis0x:
Reviewed the CLA, happy to sign.

genisis0x:
Hi @SunsetWolf — also ~7 days open. Unpickler fix that allows `Alpha158` / `Alpha360` handlers and the static-method shim to load from saved pickles after the recent module reshuffle. Would you have a moment? Thanks.

Olcmyk:
hey dude, you forget the "zscore"

genisis0x:
Thanks for double-checking. The three ZScore processor classes are already on the safelist — `ZScoreNorm`, `RobustZScoreNorm` and `CSZScoreNorm`, all under `qlib.data.dataset.processor`. The bare `zscore` / `robust_zscore` in `qlib.utils.data` are plain functions rather than classes that end up on a pickle graph, so they don't go through the RestrictedUnpickler and don't need safelisting.

If you've hit a `Forbidden class` error on a specific zscore type that isn't covered, let me know which one and I'll add it.

PR #2216, opened 2026-05-14
Fix PortAnaRecord crash on duplicate "datetime" level (#1909)

### LINKED ISSUE #1909: datetime occurs multiple times
## 🐛 Bug Description

when change example config to alpha360, datetime occurs multiple times error occurs. i debug to recognize may be record_temp.py can fix to solve

## To Reproduce

1.change the data handler to use alpha360
```
    dataset:
        class: DatasetH
        module_path: qlib.data.dataset
        kwargs:
            handler:
                class: Alpha360
                module_path: qlib.contrib.data.handler
                kwargs: *data_handler_config
            segments:
                train: [2022-04-12, 2024-08-12]
                valid: [2024-08-13, 2024-12-31]
                test: [2025-01-05, 2025-04-01]
```
2. qrun config

## Expected Behavior

no error

## Screenshot

Image

## Environment

and paste them here directly.

 - Qlib version: 0.9.6
 - Python version:3.10
 - OS (`Windows`):

## Additional Notes

after i change the code from record_temp.py function _generate
```python
    def _generate(self, **kwargs):
        pred = self.load("pred.pkl")

        # replace the "<PRED>" with prediction saved before
        placeholder_value = {"<PRED>": pred}
        for k in "executor_config", "strategy_config":
            setattr(self, k, fill_placeholder(getattr(self, k), placeholder_value))
        # if the backtesting time range is not set, it will automatically extract time range from the prediction file
        dt_values = pred.index.get_level_values("datetime")
        if self.backtest_config["start_time"] is None:
            self.backtest_config["start_time"] = dt_values.min()
        if self.backtest_config["end_time"] is None:
            self.backtest_config["end_time"] = get_date_by_shift(dt_values.max(), 1)
......
```
to 
```python
    def _generate(self, **kwargs):
        pred = self.load("pred.pkl")

        # replace the "<PRED>" with prediction saved before
        placeholder_value = {"<PRED>": pred}
        for k in "executor_config", "strategy_config":
            setattr(self, k, fill_placeholder(getattr(self, k), placeholder_value))
        # if the backtesting time range is not set, it will automatically extract time range from the prediction file
        datetime_level = pred.index.names.index("datetime")
        dt_values = pred.index.get_level_values(datetime_level)
        if self.backtest_config["start_time"] is None:
            self.backtest_config["start_time"] = dt_values.min()
        if self.backtest_config["end_time"] is None:
            self.backtest_config["end_time"] = get_date_by_shift(dt_values.max(), 1)
....
```
the error disapper

genisis0x:
## Summary

- `PortAnaRecord._generate` and `MultiPassPortAnaRecord.random_init` resolved the prediction DataFrame's `datetime` level by name via `index.get_level_values(\"datetime\")`.
- Some handler/processor chains produce a MultiIndex that carries two levels both named `datetime`. In that case pandas raises `ValueError: 'datetime' occurs multiple times`, which is the crash in #1909.
- Switch both call sites to positional lookup: `dt_level = index.names.index(\"datetime\")` then `index.get_level_values(dt_level)`. That returns the first `datetime` level deterministically and survives duplicate names.

## Test plan

- [x] New `tests/misc/test_record_temp_duplicate_datetime.py` builds a MultiIndex with duplicate `datetime` names and asserts:
  - name-based lookup raises `ValueError` (precondition)
  - positional lookup returns the expected values
- [x] Existing record-temp unit tests still pass (no behavior change on single-`datetime` indexes)

Fixes #1909

### DIFF
--- a/qlib/workflow/record_temp.py
+++ b/qlib/workflow/record_temp.py
@@ -471,7 +471,12 @@ class PortAnaRecord(ACRecordTemp):
             setattr(self, k, fill_placeholder(getattr(self, k), placeholder_value))
 
         # if the backtesting time range is not set, it will automatically extract time range from the prediction file
-        dt_values = pred.index.get_level_values("datetime")
+        # Resolve the "datetime" level positionally. With some handler chains
+        # (e.g. Alpha360 plus certain processors) the predicted DataFrame can
+        # carry two levels named "datetime", and the name-based lookup raises
+        # "datetime occurs multiple times" (#1909).
+        dt_level = pred.index.names.index("datetime")
+        dt_values = pred.index.get_level_values(dt_level)
         if self.backtest_config["start_time"] is None:
             self.backtest_config["start_time"] = dt_values.min()
         if self.backtest_config["end_time"] is None:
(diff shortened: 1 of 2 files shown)


genisis0x:

PR #2217, opened 2026-05-14
Resolve duplicate "datetime" level across remaining qlib call sites

genisis0x:
## Summary

Follow-up hardening on top of #2216 (which fixed `PortAnaRecord` for #1909).

Auditing the codebase for the same name-based `index.get_level_values(\"datetime\")` pattern surfaced several additional sites that would crash with the same `ValueError: 'datetime' occurs multiple times` on the same MultiIndex shape:

- `qlib/utils/__init__.py` — `split_pred`
- `qlib/contrib/evaluate.py` — long/short returns loop in `backtest_analysis`
- `qlib/workflow/online/update.py` — `PredUpdater.__init__` and `_replace_range`
- `qlib/workflow/online/manager.py` — `OnlineManager.prepare_signals` and `routine`
- `qlib/model/ens/ensemble.py` — `RollingEnsemble.__call__`
- `qlib/data/dataset/__init__.py` — `DatasetH.setup_data`

All call sites switch to positional level resolution:

```python
dt_level = index.names.index(\"datetime\")
index.get_level_values(dt_level)
```

This matches the resolution used in #2216.

## Test plan

- [x] New `tests/misc/test_datetime_level_lookup_helpers.py` builds the same duplicate-`datetime` MultiIndex used in the #2216 regression test and exercises `split_pred`, asserting:
  - the precondition that the name-based lookup raises
  - `split_pred` runs and returns a complete partition (left + right = input)
- [x] No behavior change on single-`datetime` indexes (positional lookup of the first `datetime` level is equivalent when the name is unique)

Refs #1909

### DIFF
--- a/qlib/contrib/evaluate.py
+++ b/qlib/contrib/evaluate.py
@@ -336,7 +336,9 @@ def long_short_backtest(
         shift=shift,
     )
 
-    _pred_dates = pred.index.get_level_values(level="datetime")
+    # Resolve positionally to survive duplicate "datetime" level names (#1909).
+    _dt_level = pred.index.names.index("datetime")
+    _pred_dates = pred.index.get_level_values(_dt_level)
     predict_dates = D.calendar(start_time=_pred_dates.min(), end_time=_pred_dates.max())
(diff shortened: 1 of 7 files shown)


genisis0x:

PR #2225, opened 2026-05-22
fix: parse mixed Yahoo date index

### LINKED ISSUE #2014: 执行yahoo normalize data 发生 ValueError: unconverted data remains when parsing with format "%Y-%m-%d" 错误
## To Reproduce

1.python scripts/data_collector/yahoo/collector.py normalize_data --source_dir ~/.qlib/stock_data/source/cn_data --normalize_dir ~/.qlib/stock_data/normalize/cn_data --region CN --interval 1d
2.我已经下载了cn_data数据，normalize 过程中，在第341  sh600489.csv 发生错误

## Screenshot
  File "E:\Data\PROGRAM\Python\source\qlib\scripts\data_collector\yahoo\collector.py", line 395, in normalize_yahoo
    df.index = pd.to_datetime(df.index)
...

ValueError: unconverted data remains when parsing with format "%Y-%m-%d": " 09:30:00+08:00", at position 1343. You might want to try:
    - passing `format` if your strings have a consistent format;
    - passing `format='ISO8601'` if your strings are all ISO8601 but not necessarily in exactly the same format;
    - passing `format='mixed'`, and the format will be inferred for each element individually. You might want to use `dayfirst` alongside this.

## Environment

and paste them here directly.

 - Qlib version:0.9.7
 - Python version: 3.12.9
 - OS (`Windows`, `Linux`, `MacOS`): windows

下载的原始数据
[sh600468.csv](https://github.com/user-attachments/files/22408114/sh600468.csv)
[sh600469.csv](https://github.com/user-attachments/files/22408113/sh600469.csv)

归一后的数据

[sh600467.csv](https://github.com/user-attachments/files/22408140/sh600467.csv)
[sh600468.csv](https://github.com/user-attachments/files/22408139/sh600468.csv)

he-yufeng:
﻿## What
- make Yahoo normalization tolerate a mixed date index containing both daily dates and timezone-bearing timestamps
- keep the existing timezone-stripping behavior after parsing
- add a small regression test for the mixed daily/timestamp shape from #2014

## To verify
- `python -m py_compile scripts\data_collector\yahoo\collector.py tests\test_yahoo_collector.py`
- `git diff --check -- scripts\data_collector\yahoo\collector.py tests\test_yahoo_collector.py`

## Local note
- `python -m pytest tests\test_yahoo_collector.py -q` is blocked in this Windows checkout because qlib's local C extensions are not built: `ModuleNotFoundError: No module named 'qlib.data._libs.rolling'`.
- `python setup.py build_ext --inplace` also fails before tests with Conda Python's `pyconfig.h` unable to include `io.h`, so I did not classify that as a patch failure.

Fixes #2014

### DIFF
--- a/scripts/data_collector/yahoo/collector.py
+++ b/scripts/data_collector/yahoo/collector.py
@@ -369,6 +369,23 @@ class YahooNormalize(BaseNormalize):
     DAILY_FORMAT = "%Y-%m-%d"
 
     @staticmethod
+    def _parse_datetime_index(index) -> pd.DatetimeIndex:
+        try:
+            parsed = pd.to_datetime(index)
+            if isinstance(parsed, pd.DatetimeIndex):
+                return parsed.tz_localize(None) if parsed.tz is not None else parsed
+        except ValueError:
+            pass
+
+        parsed_values = []
+        for value in index:
+            timestamp = pd.to_datetime(value)
+            if getattr(timestamp, "tzinfo", None) is not None:
+                timestamp = timestamp.tz_localize(None)
+            parsed_values.append(timestamp)
+        return pd.DatetimeIndex(parsed_values)
+
+    @staticmethod
     def calc_change(df: pd.DataFrame, last_close: float) -> pd.Series:
         df = df.copy()
         _tmp_series = df["close"].ffill()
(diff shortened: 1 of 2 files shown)


he-yufeng:
Quick follow-up: this is still scoped to parsing mixed Yahoo date index values without dropping valid rows, with a regression test in `tests/test_yahoo_collector.py`.

Could a maintainer take a look when convenient?

he-yufeng:
Second nudge on this one. The Yahoo collector still hardcodes separate date formats per class on current master ("%Y%m%d" for the query string, "%Y-%m-%d" as DAILY_FORMAT), so a mixed-format index from an existing CSV or a hand-built range still trips the parser. The patch keeps both accepted formats working side by side and adds a regression test for the mixed case.

PR #2237, opened 2026-05-24
fix: cache filtered instruments during grouped loads

### LINKED ISSUE #2236: [SeriesDFilter] Redundant execution of filter_pipe for each fields_group causes slow data loading
# Description
During data loading, I noticed that `filter_pipe` is executed repeatedly for each `fields_group` instead of being applied once or cached. This causes significant slowdowns in data preparation, especially when multiple field groups are configured.

# Steps to Reproduce / Observation
1. Added logging/print statements inside the `_getFilterSeries` method.
2. Ran the data loading pipeline with multiple `fields_group`s (e.g., 3 groups in my setup).
3. Observed that the filter pipeline was called exactly 3 times, once per `fields_group`, rather than sharing the filtered result across groups.

he-yufeng:
﻿## Description
`QlibDataLoader.load()` now resolves `filter_pipe` once before grouped field loading starts, then reuses the filtered instruments for each field group. Direct `load_group_df()` calls keep the old behavior by resolving instruments at that entry point.

A regression test covers grouped loading with two field groups and verifies that `D.instruments(..., filter_pipe=...)` is called once while `D.features(...)` is still called per group.

## Motivation and Context
Fixes #2236.

The previous grouped path delegated every group to `load_group_df()` with the original market name, so each group re-ran the same dynamic instrument filter pipeline. With multiple `fields_group`s, that multiplies filter work even though the filtered instrument set is identical for the load request.

## How Has This Been Tested?

Targeted checks run locally:

- `python -m py_compile qlib\data\dataset\loader.py tests\data_mid_layer_tests\test_dataloader.py`
- `git diff --check`
- an isolated loader smoke test that stubs Qlib's C extensions and asserts one `D.instruments` call plus two grouped `D.features` calls

I also attempted the targeted pytest:

- `python -m pytest tests\data_mid_layer_tests\test_dataloader.py::TestDataLoader::test_group_loader_applies_filter_pipe_once -q`

That is blocked in my Windows checkout before test collection because the local Cython extension is not built:

```text
ModuleNotFoundError: No module named 'qlib.data._libs.rolling'
```

Building the extension with `python setup.py build_ext --inplace` also fails locally because this Windows environment cannot find the Windows SDK `io.h` header while compiling `rolling.cpp`.

1. Pipeline test: not run; local C extension build is blocked as described above.
2. Your own tests: py_compile, diff check, and isolated loader smoke test passed.

## Types of changes
- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/data/dataset/loader.py
+++ b/qlib/data/dataset/loader.py
@@ -199,6 +199,35 @@ class QlibDataLoader(DLWParser):
                     self.inst_processors
                 ), f"freq(={self.freq}), inst_processors(={self.inst_processors}) cannot be None/empty"
 
+    def _resolve_instruments(self, instruments):
+        if instruments is None:
+            warnings.warn("`instruments` is not set, will load all stocks")
+            instruments = "all"
+        if isinstance(instruments, str):
+            return D.instruments(instruments, filter_pipe=self.filter_pipe)
+        if self.filter_pipe is not None:
+            warnings.warn(
+                "`filter_pipe` is not None, but it will not be used with `instruments` as list"
+            )
+        return instruments
+
+    def load(self, instruments=None, start_time=None, end_time=None) -> pd.DataFrame:
+        instruments = self._resolve_instruments(instruments)
+        if self.is_group:
+            df = pd.concat(
+                {
+                    grp: self._load_group_df(
+                        instruments, exprs, names, start_time, end_time, grp
+                    )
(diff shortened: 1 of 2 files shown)


he-yufeng:
Closing this older version in favor of #2243, which is the cleaner current patch for the same #2236 issue. Keeping only one active PR should make review easier.

PR #2240, opened 2026-05-29
fix(data/filter): apply sort_index result in SeriesDFilter

### LINKED ISSUE #2165: Bug: sort_index() result discarded in filter.py — timestamp series may be unsorted
### Description

In `qlib/data/filter.py` line 163, `timestamp_series.sort_index()` is called but its return value is discarded:

```python
timestamp_series.sort_index()  # returns new Series, original unchanged
```

`pandas.Series.sort_index()` returns a new sorted Series and does **not** modify in place (unlike `sort_index(inplace=True)`). The unsorted series is then iterated to build timestamp ranges, which can produce incorrect filter boundaries.

### Fix

```diff
- timestamp_series.sort_index()
+ timestamp_series = timestamp_series.sort_index()
```

I will submit a fix shortly.

genisis0x:
## Description

Closes #2165.

In `qlib/data/filter.py`, `SeriesDFilter._getFilterSeries` sorts the timestamp series but discards the result:

```python
# sort the timestamp_series according to the timestamps
timestamp_series.sort_index()
```

`pandas.Series.sort_index()` returns a **new** sorted Series and does not sort in place (that would require `inplace=True`). The series is therefore iterated **unsorted** in the loop that builds the `(timestamp, timestamp)` ranges, which can yield incorrect filter boundaries when the input index is not already ordered.

## Change

Assign the sorted result back: `timestamp_series = timestamp_series.sort_index()`. This matches the intent of the existing comment and the in-order assumption of the boundary scan that follows.

## Testing

With an out-of-order input index, the boundary scan now iterates ascending timestamps and produces correctly ordered ranges. Already-sorted inputs are unaffected.

### DIFF
--- a/qlib/data/filter.py
+++ b/qlib/data/filter.py
@@ -160,7 +160,7 @@ class SeriesDFilter(BaseDFilter):
             the list of tuple (timestamp, timestamp).
         """
         # sort the timestamp_series according to the timestamps
-        timestamp_series.sort_index()
+        timestamp_series = timestamp_series.sort_index()
         timestamp = []
         _lbool = None
         _ltime = None


genisis0x:
@SunsetWolf a review when convenient would be great. One-liner in SeriesDFilter: `sort_index()` return value was discarded (pandas doesn't sort in place), so the timestamp boundary scan ran on an unsorted series; now assigned back. CLA signed, CI green.

PR #2242, opened 2026-05-30
fix: compare joblib versions semantically

he-yufeng:
﻿## Summary
- compare `joblib.__version__` with `packaging.version.parse()` instead of raw strings
- add a regression test covering a future-style `1.10.0` joblib version string

## Why
`ParallelExt` selects between joblib's old `_backend_args` and newer `_backend_kwargs` attributes based on the joblib version. A raw string comparison treats `"1.10.0"` as lower than `"1.5.0"`, which would send newer joblib releases down the removed `_backend_args` path.

## Tests
- `python -m py_compile qlib\\utils\\paral.py tests\\misc\\test_paral.py`
- `python -m pytest tests\\misc\\test_paral.py -q`
- `git diff --check`

### DIFF
--- a/qlib/utils/paral.py
+++ b/qlib/utils/paral.py
@@ -9,6 +9,7 @@ from typing import Callable, Text, Union
 import joblib
 from joblib import Parallel, delayed
 from joblib._parallel_backends import MultiprocessingBackend
+from packaging import version
 import pandas as pd
 
 from queue import Empty, Queue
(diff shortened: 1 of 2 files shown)


he-yufeng:
First nudge on this one. The joblib version comparison still compares the strings lexically on current main, so 2.10 sorts below 2.9; this makes it a semantic compare with a regression test. The branch is clean against main right now, so review should be cheap.

he-yufeng:
This one is still current: paral.py:27 still compares joblib.__version__ < "1.5.0" as a plain string, so the check misbehaves again once minor versions hit two digits. Could a maintainer take another pass?

PR #2243, opened 2026-05-31
fix: reuse filtered instruments across field groups

### LINKED ISSUE #2236: [SeriesDFilter] Redundant execution of filter_pipe for each fields_group causes slow data loading

he-yufeng:
﻿Fixes #2236.

`QlibDataLoader` resolves `filter_pipe` inside `load_group_df()`. With grouped field configs, `load()` calls `load_group_df()` once per fields group, so the same market string runs through `D.instruments(..., filter_pipe=...)` repeatedly.

This resolves string instruments once at the grouped-loader entry point and reuses that instrument config for each group. Direct `load_group_df()` calls keep the old behavior, and explicit non-string instrument lists still warn that `filter_pipe` is ignored.

## To verify

- `python -m pytest tests\data_mid_layer_tests\test_dataloader.py -q -k grouped_loader`
- `python -m py_compile qlib\data\dataset\loader.py tests\data_mid_layer_tests\test_dataloader.py`
- `git diff --check`

### DIFF
--- a/qlib/data/dataset/loader.py
+++ b/qlib/data/dataset/loader.py
@@ -199,6 +199,11 @@ class QlibDataLoader(DLWParser):
                     self.inst_processors
                 ), f"freq(={self.freq}), inst_processors(={self.inst_processors}) cannot be None/empty"
 
+    def load(self, instruments=None, start_time=None, end_time=None) -> pd.DataFrame:
+        if self.is_group and isinstance(instruments, str):
+            instruments = D.instruments(instruments, filter_pipe=self.filter_pipe)
+        return super().load(instruments, start_time, end_time)
+
     def load_group_df(
         self,
         instruments,
@@ -213,7 +218,7 @@ class QlibDataLoader(DLWParser):
             instruments = "all"
         if isinstance(instruments, str):
             instruments = D.instruments(instruments, filter_pipe=self.filter_pipe)
-        elif self.filter_pipe is not None:
+        elif self.filter_pipe is not None and not isinstance(instruments, dict):
             warnings.warn("`filter_pipe` is not None, but it will not be used with `instruments` as list")
 
         freq = self.freq[gp_name] if isinstance(self.freq, dict) else self.freq
--- a/qlib/data/dataset/loader.py
+++ b/qlib/data/dataset/loader.py
@@ -200,9 +200,42 @@ class QlibDataLoader(DLWParser):
                 ), f"freq(={self.freq}), inst_processors(={self.inst_processors}) cannot be None/empty"
 
     def load(self, instruments=None, start_time=None, end_time=None) -> pd.DataFrame:
-        if self.is_group and isinstance(instruments, str):
-            instruments = D.instruments(instruments, filter_pipe=self.filter_pipe)
+        if self.is_group:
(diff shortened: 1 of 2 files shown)


he-yufeng:
Updated this with an Alpha158 handler-level regression test as well. The new test constructs `Alpha158(..., filter_pipe=..., infer_processors=[], learn_processors=[])` and checks that `D.instruments(..., filter_pipe=...)` is still called once while feature and label groups are loaded separately.

Local checks:

- `python -m pytest tests/data_mid_layer_tests/test_dataloader.py::TestDataLoader::test_grouped_loader_applies_filter_pipe_once tests/data_mid_layer_tests/test_dataloader.py::TestDataLoader::test_alpha158_handler_applies_filter_pipe_once -q` -> 2 passed
- `python -m py_compile qlib\data\dataset\loader.py tests\data_mid_layer_tests\test_dataloader.py`
- `git diff --check`

I also ran the whole `tests/data_mid_layer_tests/test_dataloader.py` file after installing the missing `mlflow` dependency. The two filter-pipe tests pass, but the pre-existing `test_nested_data_loader` still needs local Qlib market data under `~/.qlib/qlib_data/cn_data` and fails before reaching this change.

he-yufeng:
Thanks for the Alpha158 repro detail. The first version only moved `D.instruments(...)` out of the per-group loop, but `D.features(...)` still resolves stockpool configs through `list_instruments(...)` for each group, so feature and label could still apply the dynamic filter twice.

I pushed a follow-up commit that resolves a stockpool config with `filter_pipe` to the filtered instrument dict once per grouped loader/frequency, then reuses that dict for each group. The regression tests now assert that `D.list_instruments(...)` is called once and both feature/label groups receive the resolved instruments.

Local checks:

- `python -m pytest tests/data_mid_layer_tests/test_dataloader.py::TestDataLoader::test_grouped_loader_applies_filter_pipe_once tests/data_mid_layer_tests/test_dataloader.py::TestDataLoader::test_alpha158_handler_applies_filter_pipe_once -q` -> 2 passed
- `python -m py_compile qlib\data\dataset\loader.py tests\data_mid_layer_tests\test_dataloader.py`
- `git diff --check`

I also reran the full `tests/data_mid_layer_tests/test_dataloader.py`; the two relevant tests pass, while the pre-existing `test_nested_data_loader` still fails locally because this Windows machine does not have usable day-frequency Qlib market data under `C:\Users\He\.qlib\qlib_data\cn_data`.

he-yufeng:
Still applies on current main: DLWParser.load still calls load_group_df once per group, and the dynamic filter is still re-applied on every call. Whenever someone has a moment, another look at this would be great.

Issue #1854, opened 2024-10-24
An error when I do 30min trade.


DOCCA0:
_ValueError: can't find a freq from [Freq(30min)] that can resample to 1min!_

code:
```python

###################################
# prediction, backtest & analysis
###################################
port_analysis_config = {
    "executor": {
        "class": "SimulatorExecutor",
        "module_path": "qlib.backtest.executor",
        "kwargs": {
            "time_per_step": "30min",
            "generate_portfolio_metrics": True,
        },
    },
    "strategy": {
        "class": "TopkDropoutStrategy",
        "module_path": "qlib.contrib.strategy.signal_strategy",
        "kwargs": {
            "model": model,
            "dataset": dataset,
            "topk": 10,
            "n_drop": 5,
        },
    },
    "backtest": {
        "start_time": "2024-08-01 00:00:00",
        "end_time": "2024-10-30 23:59:59",
        "account": 10000,
        "benchmark": benchmark,
        "exchange_kwargs": {
            "freq": "30min",
            "limit_threshold": 0.095,
            "deal_price": "close",
            "open_cost": 0.0005,
            "close_cost": 0.0015,
            "min_cost": 5,
        },
    },
}

# backtest and analysis
with R.start(experiment_name="backtest_analysis"):
    recorder = R.get_recorder(recorder_id=rid, experiment_name="train_model")
    model = recorder.load_object("trained_model")

    # prediction
    recorder = R.get_recorder()
    ba_rid = recorder.id
    sr = SignalRecord(model, dataset, recorder)
    sr.generate()

    # backtest & analysis
    par = PortAnaRecord(recorder, port_analysis_config, "day")
    par.generate()
```

PR #2247, opened 2026-06-04
fix: handle missing minute-level data in get_higher_eq_freq_feature

Whning0513:
This was the most involved fix of the three I worked on. #1854 causes a backtest crash when `time_per_step='30min'` is used — the error message says `ValueError: can't find a freq from [Freq(30min)] that can resample to 1min!`, which is confusing because the user never asked for 1min data.

I traced the crash from the backtest engine down through the call chain. `PortfolioMetrics._cal_benchmark()` calls `D.features()` with the 30min frequency. Since the CSI300 benchmark only has daily data, a ValueError is raised. The code then enters the minute-level fallback branch in `get_higher_eq_freq_feature()` (`qlib/utils/resam.py`), which hardcodes a retry with `freq='1min'` — but this fallback is **not wrapped in try/except**, unlike the day-level fallback earlier in the same function. When 1min data also doesn't exist, the raw exception propagates up as the misleading error above.

While reading the code, I also found a type inconsistency in `Freq.get_recent_freq()` (`qlib/utils/time.py`). The method builds a list of candidate frequencies — the first entry was stored as `str(_freq)` while subsequent entries kept the raw `_freq` (which could be either `str` or `Freq`). This made the return type unpredictable for callers like `FileCalendarStorage._freq_file`.

The fix spans two files:

1. **`qlib/utils/resam.py`** — Added a `try/except (ValueError, KeyError)` around the `1min` fallback inside `get_higher_eq_freq_feature()`, matching the pattern already used by the day/week/month branch a few lines above. If 1min also fails, it falls through to `day` as the ultimate fallback.

2. **`qlib/utils/time.py`** — Wrapped all candidates in `Freq.get_recent_freq()` consistently with `Freq(_freq)`, so the return type is always `Freq | None` regardless of which candidate wins.

Would appreciate a review, thanks!

### DIFF
--- a/qlib/utils/resam.py
+++ b/qlib/utils/resam.py
@@ -92,8 +92,12 @@ def get_higher_eq_freq_feature(instruments, fields, start_time=None, end_time=No
                 _result = D.features(instruments, fields, start_time, end_time, freq="1min", disk_cache=disk_cache)
                 _freq = "1min"
         elif norm_freq == Freq.NORM_FREQ_MINUTE:
-            _result = D.features(instruments, fields, start_time, end_time, freq="1min", disk_cache=disk_cache)
-            _freq = "1min"
+            try:
+                _result = D.features(instruments, fields, start_time, end_time, freq="1min", disk_cache=disk_cache)
+                _freq = "1min"
+            except (ValueError, KeyError):
+                _result = D.features(instruments, fields, start_time, end_time, freq="day", disk_cache=disk_cache)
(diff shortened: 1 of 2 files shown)


Whning0513:

Whning0513:

Whning0513:

Whning0513:

Whning0513:

Whning0513:

PR #2047, opened 2025-11-06
fix(data_collector): fix us_index collector.py Http Error 403 Forbidden; Remove FutureWarning

kzhdev:
## Description
1. Fix us_index collector.py Http Error: 403 Forbidden
2. Reomve FutureWarning

## Motivation and Context
us_index collector.py stopped working which caused by the following error:

Traceback (most recent call last):
  File "Y:\repo\qlib\scripts\data_collector\us_index\collector.py", line 273, in <module>
    fire.Fire(partial(get_instruments, market_index="us_index"))
  File "C:\Users\auror\miniforge3\envs\qlib\Lib\site-packages\fire\core.py", line 135, in Fire
    component_trace = _Fire(component, args, parsed_flag_args, context, name)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\auror\miniforge3\envs\qlib\Lib\site-packages\fire\core.py", line 559, in _Fire
    component, remaining_args = _CallAndUpdateTrace(
                                ^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\auror\miniforge3\envs\qlib\Lib\site-packages\fire\core.py", line 684, in _CallAndUpdateTrace
    component = fn(*varargs, **kwargs)
                ^^^^^^^^^^^^^^^^^^^^^^
  File "\\Trade-server\d\repo\qlib\scripts\data_collector\utils.py", line 672, in get_instruments
    getattr(obj, method)()
  File "\\Trade-server\d\repo\qlib\scripts\data_collector\index.py", line 213, in parse_instruments
    changers_df = self.get_changes()
                  ^^^^^^^^^^^^^^^^^^
  File "\\Trade-server\d\repo\qlib\scripts\data_collector\us_index\collector.py", line 229, in get_changes
    changes_df = pd.read_html(self.WIKISP500_CHANGES_URL)[-1]
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\auror\miniforge3\envs\qlib\Lib\site-packages\pandas\io\html.py", line 1240, in read_html
    return _parse(
           ^^^^^^^
  File "C:\Users\auror\miniforge3\envs\qlib\Lib\site-packages\pandas\io\html.py", line 983, in _parse
    tables = p.parse_tables()
             ^^^^^^^^^^^^^^^^
[...]

PR #2258, opened 2026-06-16
fix(cn_index): wrap literal HTML in StringIO for pd.read_html

he-yufeng:
## Description

#2047 wrapped the literal HTML passed to `pd.read_html` in a `StringIO` for the `us_index` collector to silence pandas' `FutureWarning`:

> Passing literal html to 'read_html' is deprecated and will be removed in a future version. To read from a literal string, wrap it in a 'StringIO' object.

The parallel `cn_index` collector was missed. `_parse_table` passes the csindex response body (an HTML string, not a URL/path) straight to `pd.read_html(content)` at `scripts/data_collector/cn_index/collector.py:185`, so it still hits the same deprecation and will break once pandas drops literal-string support.

## Change

Import `StringIO` and wrap the content, matching what #2047 did for `us_index`. No behavior change on supported pandas versions; `StringIO` has always been accepted by `read_html`.

Reproduced locally on pandas 2.3.3:

```
>>> pd.read_html("<table>...</table>")        # FutureWarning: Passing literal html ...
>>> pd.read_html(StringIO("<table>...</table>"))  # no warning
```

## Note

I scoped this to the `read_html` fix to stay a clean parallel of #2047. The same file also calls `DataFrame.applymap` (line 174), which pandas 2.1 deprecates in favour of `DataFrame.map`, but `DataFrame.map` only exists from 2.1 onward while `setup.py` still supports `pandas>=1.1`, so swapping it would need version handling and is left out here.

### DIFF
--- a/scripts/data_collector/cn_index/collector.py
+++ b/scripts/data_collector/cn_index/collector.py
@@ -4,7 +4,7 @@
 import re
 import abc
 import sys
-from io import BytesIO
+from io import BytesIO, StringIO
 from typing import List, Iterable
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


he-yufeng:
Gentle bump on this one. It wraps the literal HTML fragment in a String container so the cn_index reader stops tripping over it. Small and self-contained. Happy to rebase if it has drifted, or adjust anything you would prefer.

he-yufeng:
Second gentle bump. scripts/data_collector/cn_index/collector.py is unchanged since July: pd.read_html still gets the raw HTML literal, deprecated on current pandas and removed in pandas 3. The StringIO wrap here keeps both working.

PR #2262, opened 2026-06-18
fix(utils): strip the _qlib_ prefix in fname_to_code, not characters

he-yufeng:
## Description

`fname_to_code()` is meant to reverse `code_to_fname()`, which prefixes reserved Windows device names (`CON`, `PRN`, `AUX`, `NUL`, `COM0`–`COM9`, `LPT0`–`LPT9`) with `_qlib_` so they can be used as directory names. The reverse used `fname.lstrip("_qlib_")`, but `str.lstrip` removes *any* leading characters contained in its argument, not the prefix as a whole.

`exists_qlib_data()` lowercases each feature directory name before calling `fname_to_code()`:

```python
code_names = set(map(lambda x: fname_to_code(x.name.lower()), features_dir.iterdir()))
```

So `LPT1` is stored as `_qlib_LPT1`, lowercased to `_qlib_lpt1`, and then `lstrip("_qlib_")` strips the leading `l` too (it is one of the prefix characters), yielding `pt1` instead of `lpt1`. The data-existence check then fails to match such a stock. Any name whose body begins with a character in `{_, q, l, i, b}` is corrupted (e.g. `fname_to_code("_qlib_lll")` returns `""`).

The fix removes the prefix by slicing, so the original code is recovered intact.

## Verification

Before:

```python
>>> fname_to_code("_qlib_lll")
''
>>> fname_to_code(code_to_fname("LPT1").lower())
'pt1'
```

After:

```python
>>> fname_to_code("_qlib_lll")
'lll'
>>> fname_to_code(code_to_fname("LPT1").lower())
'lpt1'
```

Added `FileNameUtils.test_fname_code_round_trip` in `tests/misc/test_utils.py`, covering the reserved device names (via the lowercased path `exists_qlib_data` uses), a prefix-only body, and a plain code. `black -l 120 --check` passes on the changed files.

### DIFF
--- a/qlib/utils/__init__.py
+++ b/qlib/utils/__init__.py
@@ -932,7 +932,10 @@ def fname_to_code(fname: str):
 
     prefix = "_qlib_"
     if fname.startswith(prefix):
-        fname = fname.lstrip(prefix)
+        # NOTE: use slicing rather than ``lstrip(prefix)``; ``str.lstrip`` strips
+        # any leading characters contained in ``prefix`` (e.g. it would turn
+        # "_qlib_lpt1" into "pt1"), not the prefix as a whole.
+        fname = fname[len(prefix) :]
     return fname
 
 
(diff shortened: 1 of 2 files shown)


he-yufeng:
Following up on this. It strips the stray _qlib_ prefix that leaks into feature names. Low risk, one spot. Would appreciate a review whenever you get the chance.

he-yufeng:
Second bump. qlib/utils/__init__.py has not moved since July, so fname_to_code still keeps the _qlib_ prefix instead of stripping it. Anything you want changed here?

PR #2281, opened 2026-06-23
Fix TopkDropoutStrategy dynamic risk sizing

### LINKED ISSUE #2276: TopkDropoutStrategy ignores get_risk_degree() for position sizing
## 🐛 Bug Description

`get_risk_degree()` is documented as the dynamic market-timing hook ("Dynamically risk_degree will result in Market timing.", `qlib/contrib/strategy/signal_strategy.py:69`). But `TopkDropoutStrategy.generate_trade_decision` sizes new buys with the **raw attribute** `self.risk_degree`, not the getter:

```python
# qlib/contrib/strategy/signal_strategy.py:266
value = cash * self.risk_degree / len(buy) if len(buy) > 0 else 0
```

It is the **only** use of risk_degree in `TopkDropoutStrategy`, and it bypasses `get_risk_degree()`. So a subclass that overrides `get_risk_degree()` to return a time-varying multiplier (the documented way to implement market timing) has **zero effect** on a TopkDropout book — the override is never called. (By contrast, `WeightStrategyBase` at line 365 and its subclass `EnhancedIndexingStrategy` at line 482 *do* call `self.get_risk_degree(...)`, so the inconsistency is within the same module.)

## To Reproduce

1. Subclass `TopkDropoutStrategy` and override `get_risk_degree()` to return, say, `0.0` on some dates (a regime gate to cash) and `self.risk_degree` otherwise.
2. Backtest it vs. the un-subclassed `TopkDropoutStrategy` on the same signal.
3. The two produce **identical** trades/returns — the overridden `get_risk_degree()` is never consulted for buy sizing.

## Expected Behavior

`TopkDropoutStrategy` should size buys with `self.get_risk_degree(trade_step)` (as `WeightStrategyBase` does), so that overriding `get_risk_degree()` actually enables market timing as documented:

```python
value = cash * self.get_risk_degree(trade_step) / len(buy) if len(buy) > 0 else 0
```

## Environment

- Qlib version: 0.9.7 (`qlib/contrib/strategy/signal_strategy.py` (`TopkDropoutStrategy.generate_trade_decision`, line ~266)
- Python version: 3.12

## Additional context

Discovered while building a BTC-trend regime overlay (`RegimeGatedTopkStrategy(TopkDropoutStrategy)`) that gates gross exposure via `get_risk_degree()`. The overlay had no effect on backtests until we worked around it by pushing the gated value onto `self.risk_degree` per step before delegating to `super().generate_trade_decision()`. A two-line fix in `TopkDropoutStrategy` (use the getter) would remove the need for the workaround.

yinli-systems:
## Summary

Fixes #2276.

`TopkDropoutStrategy` exposes `get_risk_degree(trade_step)` so subclasses can implement dynamic market-timing / position-sizing logic, but the buy budget used the fixed `self.risk_degree` field directly. This makes dynamic risk overrides ineffective for new buys.

This PR routes the buy budget through `get_risk_degree(trade_step)` and adds a focused regression test that keeps `self.risk_degree = 0.95` while overriding `get_risk_degree()` to return `0.25`; the generated buy order must use the dynamic 25% budget.

## Tests

- `PYTHONPYCACHEPREFIX=/private/tmp/qlib-pycache /usr/bin/python3 -m py_compile qlib/contrib/strategy/signal_strategy.py tests/backtest/test_topk_dropout_strategy.py`
- `git diff --check`
- `PYTHONPYCACHEPREFIX=/private/tmp/qlib-pycache /opt/homebrew/bin/python3.11 - <<'PY' ...` isolated harness that stubs Qlib package-level imports, loads `qlib/contrib/strategy/signal_strategy.py`, and verifies dynamic risk produces a 25-share order from 1000 cash at price 10

Note: direct local `pytest -q tests/backtest/test_topk_dropout_strategy.py` is blocked in this unbuilt source tree by missing compiled extensions (`qlib.data._libs.rolling`). I attempted `setup.py build_ext --inplace`; this checkout lacks generated C++ sources and needs the project Cython build path first. The added test is a normal repository test and should run in CI after the extension build step.

### DIFF
--- a/qlib/contrib/strategy/signal_strategy.py
+++ b/qlib/contrib/strategy/signal_strategy.py
@@ -263,7 +263,7 @@ class TopkDropoutStrategy(BaseSignalStrategy):
         # buy new stock
         # note the current has been changed
         # current_stock_list = current_temp.get_stock_list()
-        value = cash * self.risk_degree / len(buy) if len(buy) > 0 else 0
+        value = cash * self.get_risk_degree(trade_step) / len(buy) if len(buy) > 0 else 0
 
         # open_cost should be considered in the real trading environment, while the backtest in evaluate.py does not
         # consider it as the aim of demo is to accomplish same strategy as evaluate.py, so comment out this line
--- a/tests/backtest/test_topk_dropout_strategy.py
+++ b/tests/backtest/test_topk_dropout_strategy.py
@@ -1,0 +1,78 @@
+import pandas as pd
+
+from qlib.backtest.decision import Order
+from qlib.backtest.position import Position
+from qlib.contrib.strategy.signal_strategy import TopkDropoutStrategy
+
+
+class MockInfra:
+    def __init__(self, values):
+        self.values = values
+
+    def get(self, name):
+        return self.values[name]
+
+
+class MockAccount:
+    def __init__(self, position):
+        self.current_position = position
+
+
+class MockCalendar:
+    def get_trade_step(self):
+        return 0
+
+    def get_step_time(self, trade_step=None, shift=0):
+        if shift:
+            return pd.Timestamp("2020-01-01"), pd.Timestamp("2020-01-01")
+        return pd.Timestamp("2020-01-02"), pd.Timestamp("2020-01-02")
+
+
+class MockSignal:
+    def get_signal(self, start_time, end_time):
+        return pd.Series({"A": 1.0})
+
+
+class MockExchange:
+    def is_stock_tradable(self, stock_id, start_time, end_time, direction=None):
+        return True
+
+    def get_deal_price(self, stock_id, start_time, end_time, direction):
+        return 10.0
+
+    def get_factor(self, stock_id, start_time, end_time):
+        return 1.0
+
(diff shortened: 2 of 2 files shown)


yinli-systems:
Quick follow-up on scope: this is intentionally a small fix for #2276. `TopkDropoutStrategy` already exposes `get_risk_degree(trade_step)` for dynamic risk / market timing, and `WeightStrategyBase` already routes sizing through that hook. This PR makes the TopK dropout buy budget use the same hook instead of the fixed `self.risk_degree` field.

I added a focused regression test where `self.risk_degree` remains 0.95 but an override returns 0.25, and the generated buy order is sized from the dynamic 25% budget. Happy to adjust the test style if maintainers prefer this covered in an existing backtest test file.

yinli-systems:
Closing as a duplicate in favor of the earlier #2277. It makes the same one-line hook correction and includes a full Qlib regression showing an overridden risk degree gates buys while preserving the default strategy path. There is no useful reason to keep parallel implementations open.

PR #2286, opened 2026-06-29
Use dynamic risk degree in TopkDropoutStrategy

### LINKED ISSUE #2276: TopkDropoutStrategy ignores get_risk_degree() for position sizing

Whning0513:
Fixes #2276

## Summary
- size TopkDropoutStrategy buys with `get_risk_degree(trade_step)` instead of the raw `risk_degree` attribute
- preserve the existing behavior for strategies that do not override `get_risk_degree`
- add a regression test that exercises the buy-sizing path with a dynamic risk overlay

## Testing
- python -m unittest tests.backtest.test_topk_dropout_risk_degree

### DIFF
--- a/qlib/contrib/strategy/signal_strategy.py
+++ b/qlib/contrib/strategy/signal_strategy.py
@@ -263,7 +263,7 @@ class TopkDropoutStrategy(BaseSignalStrategy):
         # buy new stock
         # note the current has been changed
         # current_stock_list = current_temp.get_stock_list()
-        value = cash * self.risk_degree / len(buy) if len(buy) > 0 else 0
+        value = cash * self.get_risk_degree(trade_step) / len(buy) if len(buy) > 0 else 0
 
         # open_cost should be considered in the real trading environment, while the backtest in evaluate.py does not
         # consider it as the aim of demo is to accomplish same strategy as evaluate.py, so comment out this line
--- a/tests/backtest/test_topk_dropout_risk_degree.py
+++ b/tests/backtest/test_topk_dropout_risk_degree.py
@@ -1,0 +1,178 @@
+import unittest
+from importlib.util import module_from_spec, spec_from_file_location
+from pathlib import Path
+from types import ModuleType, SimpleNamespace
+from unittest.mock import patch
+
+import pandas as pd
+
+
+class _DummyPosition:
+    def __init__(self, cash=100.0):
+        self._cash = cash
+
+    def get_cash(self):
+        return self._cash
+
+    def get_stock_list(self):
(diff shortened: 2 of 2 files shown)


Whning0513:
Closing this in favor of the earlier PRs #2277 and #2281, which address the same issue. Thanks.

PR #2298, opened 2026-07-09
fix(ops): guard Rsquare expanding (N==0) path against degenerate windows

### LINKED ISSUE #2297: Bug: Rsquare(N=0) expanding path leaks inf/garbage on near-constant windows (rolling path is guarded)
### Description

`Rsquare` in `qlib/data/ops.py` computes R² via a Cython kernel as
`num / sqrt(var_x * var_y)`. For a **near-constant** window `var_y ≈ 0`, so
floating-point cancellation yields `inf` or a spurious finite value instead of
`NaN` (a degenerate 0/0 regression).

`Rsquare._load_internal` guards against this by masking windows whose std is ≈0
to `NaN` — but **only on the rolling (`N != 0`) branch**:

```python
    _series = self.feature.load(instrument, start_index, end_index, *args)
    if self.N == 0:
        series = pd.Series(expanding_rsquare(_series.values), index=_series.index)
        # <-- no guard here
    else:
        series = pd.Series(rolling_rsquare(_series.values, self.N), index=_series.index)
        series.loc[np.isclose(_series.rolling(self.N, min_periods=1).std(), 0, atol=2e-05)] = np.nan
    return series
```

The expanding (`N == 0`) branch is unguarded, so `Rsquare($feature, 0)` returns
`inf`/garbage on near-constant windows. Because `ops.py` sets
`np.seterr(invalid="ignore")`, no warning is emitted — the bad values silently
propagate into features (e.g. Alpha158/Alpha360) and downstream models.

### Reproduction

Near-constant series `[100, 100, 100, 100.000001, 100, 100]`:

```
expanding_rsquare (N==0 path): [nan, nan, nan, inf, 0.01717987, inf]
rolling_rsquare(4) after mask: [nan, nan, nan, nan, nan, nan]
```

The expanding path leaks `inf` and a spurious `0.0172`; the rolling path is
correctly `NaN`.

### Fix

Apply the same std≈0 → `NaN` mask on the expanding branch (using expanding std).
`Slope`/`Resi` are unaffected — they divide by the x-variance (index `1..N`),
which is always well-conditioned; only `Rsquare` divides by the y-variance.

PR incoming.

DogInfantry:
Fixes #2297

### Problem

`Rsquare` masks windows whose std is ≈0 to `NaN` so a degenerate 0/0 regression
(near-constant window) does not leak `inf`/garbage R² out of the Cython kernel.
That guard was only on the rolling (`N != 0`) branch; the expanding (`N == 0`)
branch was unguarded, so `Rsquare($feature, 0)` returned `inf`/spurious finite
values on near-constant windows. `qlib/data/ops.py` also sets
`np.seterr(invalid="ignore")`, so the bad values propagate silently into
features (Alpha158/Alpha360) and models.

### Fix

Mirror the existing guard on the expanding branch, using expanding std:

```python
if self.N == 0:
    series = pd.Series(expanding_rsquare(_series.values), index=_series.index)
    series.loc[np.isclose(_series.expanding(min_periods=1).std(), 0, atol=2e-05)] = np.nan
else:
    ...
```

`Slope`/`Resi` are unaffected — they divide by the x-variance (index `1..N`),
which is always well-conditioned; only `Rsquare` divides by the y-variance.

### Test

Adds `tests/ops/test_rsquare.py` exercising `Rsquare._load_internal` directly
(no market data needed). It **fails before** this change and **passes after**:

```
# before fix, expanding path:
[nan, nan, nan, inf, 0.01717987, inf]   -> AssertionError: expanding Rsquare leaked inf
# after fix:
3 passed
```

### DIFF
--- a/qlib/data/ops.py
+++ b/qlib/data/ops.py
@@ -1277,6 +1277,7 @@ class Rsquare(Rolling):
         _series = self.feature.load(instrument, start_index, end_index, *args)
         if self.N == 0:
             series = pd.Series(expanding_rsquare(_series.values), index=_series.index)
+            series.loc[np.isclose(_series.expanding(min_periods=1).std(), 0, atol=2e-05)] = np.nan
         else:
             series = pd.Series(rolling_rsquare(_series.values, self.N), index=_series.index)
             series.loc[np.isclose(_series.rolling(self.N, min_periods=1).std(), 0, atol=2e-05)] = np.nan
--- a/tests/ops/test_rsquare.py
+++ b/tests/ops/test_rsquare.py
@@ -1,0 +1,60 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+
+import unittest
+
+import numpy as np
+import pandas as pd
+
+from qlib.data.ops import Rsquare
+
+
+class _StubFeature:
+    """Minimal Expression stub: returns a fixed series from ``load`` so that
+    Rolling operators can be exercised without a data provider / market data."""
+
+    def __init__(self, series: pd.Series):
+        self._series = series
+
+    def load(self, *args, **kwargs) -> pd.Series:
+        return self._series
+
+
+class TestRsquareDegenerateWindow(unittest.TestCase):
+    """Regression tests for Rsquare on (near-)constant windows.
+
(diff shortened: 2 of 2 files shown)


DogInfantry:

PR #2300, opened 2026-07-10
fix(data): sandbox feature-expression parser to prevent code execution (CWE-94)

andesyteoss:
## Description

`ExpressionProvider.get_expression` and `DiskExpressionCache._uri` evaluate feature-expression strings with the builtin `eval`:

```python
# qlib/data/data.py (before)
expression = eval(parse_field(field))

# qlib/data/cache.py (before)
if not isinstance(eval(parse_field(field)), Feature):
    ...
```

`parse_field` (in `qlib/utils/__init__.py`) only performs regex rewrites — it turns `$close` into `Feature("close")` and `Ref(...)` into `Operators.Ref(...)`, but it does **not** validate or sanitize the input. Any other Python syntax in the field string is passed straight through to `eval`, which happily executes it in a scope that has `__builtins__` available. That is a textbook CWE-94 (Improper Control of Generation of Code / code injection) sink.

Field strings reach these functions from user-controlled workflow YAML configs (`qrun config.yaml`) and from dataset/handler configs shared between users — a very common qlib workflow. A malicious config such as

```yaml
data_handler_config:
  fields: ["__import__('os').system('touch /tmp/pwned')"]
```

executes arbitrary code the moment the handler is initialized.

## Motivation and Context

CWE-94 — arbitrary code execution via feature expression strings. Qlib's ecosystem revolves around sharing YAML configs (examples in `examples/`, community strategies, tutorials), so "the attacker controls a field string" is a realistic threat, not a theoretical one. Opening a shared config should not be equivalent to running `python -c "$(cat config.yaml)"`.

## Fix

This PR replaces `eval(parse_field(field))` with a new `parse_expression(field)` helper in `qlib/data/expression_parser.py`. The helper:

1. Runs the existing `parse_field` regex rewrite (so the DSL still works identically).
2. Parses the resulting string with `ast.parse(..., mode="eval")`.
3. Walks the AST and rejects anything outside a strict whitelist:
   - Literals (`ast.Constant`, plus legacy `Num`/`Str`/`NameConstant`), tuples/lists.
   - Unary `+`/`-`, binary arithmetic, boolean ops, comparisons.
   - `Name` nodes only if they resolve to `Feature`, `PFeature`, or `Operators`.
   - `Attribute` nodes only on `Operators` and only for non-dunder names.
   - `Call` nodes with positional and simple keyword arguments (no `*args`/`**kwargs` unpacking).
4. Evaluates the whitelisted tree with `__builtins__` cleared and a fixed name table `{Feature, PFeature, Operators}`.

Attribute access, subscripting, comprehensions, lambdas, f-strings, `__import__`, dunder traversal, and any unknown identifier all raise `ValueError` before any code runs.

Call sites in `qlib/data/data.py` and `qlib/data/cache.py` switch from `eval(parse_field(...))` to `parse_expression(...)`. The public `parse_field` helper is left in place for backward compatibility with any downstream users that only want the string rewrite.

## How Has This Been Tested?

- [x] Added `tests/misc/test_expression_parser_safety.py` with cases covering:
  - Legitimate expressions (`$close`, `Ref($close, -1)`, `Mean($high-$low, 5) / $close`) parse to the same objects the old `eval` path produced.
  - Malicious payloads are rejected: `__import__('os').system('id')`, `().__class__.__bases__[0].__subclasses__()`, `(lambda: 1)()`, `[x for x in range(1)]`, `open('/etc/passwd')`, `Operators.__class__`, `Ref($close, **{'d': -1})`.
- [x] Ran the new test file locally — all pass.
- [x] Manually re-ran a handful of existing handler configs from `examples/benchmarks/` through `parse_expression` to confirm parity with the previous `eval` behavior.

## Security analysis

We verified this is exploitable end-to-end: constructing a `QlibDataLoader` (or running `qrun`) with a `fields` entry containing `__import__('os').system('...')` executes the payload on the current interpreter, with the current user's privileges, at handler-initialization time — no other preconditions.

The fix removes the `eval` sink entirely. The AST walker refuses any node type it hasn't explicitly whitelisted, so future additions to Python's grammar can't silently widen the attack surface — a new node type will raise `ValueError` until it's reviewed and added. Attribute access is restricted to `Operators.<name>` with a dunder check, which blocks the classic `().__class__.__bases__[0].__subclasses__()` escape.

### Adversarial review

Before submitting, we tried to disprove this. We checked whether qlib treats configs as trusted-by-default (it doesn't — configs are routinely shared and downloaded), whether any upstream layer sanitizes field strings before they reach `eval` (no — `parse_field` is purely a regex rewrite), and whether the attacker needs privileges that would already grant equivalent access (no — reading a YAML file is not equivalent to arbitrary Python execution). We also confirmed the whitelist doesn't over-restrict: every operator in `qlib/data/ops.py` is reachable via `Operators.<name>(...)`, which the whitelist permits.

## Proof of Concept

```python
# poc.py — run from a qlib checkout with a data provider configured
from qlib.data.data import ExpressionProvider

class P(ExpressionProvider):
    def expression(self, *a, **k): pass

# Before the patch, this executes os.system('touch /tmp/qlib_pwned'):
P().get_expression_instance("__import__('os').system('touch /tmp/qlib_pwned')")
```

Or via a workflow config:

```yaml
# pwn.yaml — passed to `qrun pwn.yaml`
task:
  dataset:
    kwargs:
      handler:
        kwargs:
          data_loader:
            kwargs:
              config:
                feature: ["__import__('os').system('touch /tmp/qlib_pwned')"]
```

After this patch both raise `ValueError: unsupported expression element ...` and no code executes.

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

### DIFF
--- a/qlib/data/cache.py
+++ b/qlib/data/cache.py
@@ -25,7 +25,6 @@ from ..utils import (
     hash_args,
     get_redis_connection,
     read_bin,
-    parse_field,
     remove_fields_space,
     normalize_cache_fields,
     normalize_cache_instruments,
@@ -34,6 +33,7 @@ from ..utils.pickle_utils import restricted_pickle_load
 
 from ..log import get_module_logger
 from .base import Feature
+from .expression_parser import parse_expression
 from .ops import Operators  # pylint: disable=W0611  # noqa: F401
 
 
@@ -540,7 +540,7 @@ class DiskExpressionCache(ExpressionCache):
             field = remove_fields_space(field)
             # cache unavailable, generate the cache
             _instrument_dir.mkdir(parents=True, exist_ok=True)
-            if not isinstance(eval(parse_field(field)), Feature):
+            if not isinstance(parse_expression(field), Feature):
                 # When the expression is not a raw feature
                 # generate expression cache if the feature is not a Feature
                 # instance
--- a/qlib/data/data.py
+++ b/qlib/data/data.py
@@ -28,7 +28,6 @@ from ..utils import (
     init_instance_by_config,
     register_wrapper,
     get_module_by_module_path,
-    parse_field,
     hash_args,
     normalize_cache_fields,
     code_to_fname,
@@ -38,6 +37,7 @@ from ..utils import (
 )
 from ..utils.paral import ParallelExt
 from .ops import Operators  # pylint: disable=W0611  # noqa: F401
+from .expression_parser import parse_expression
 
 
 class ProviderBackendMixin:
@@ -394,7 +394,7 @@ class ExpressionProvider(abc.ABC):
             if field in self.expression_instance_cache:
                 expression = self.expression_instance_cache[field]
             else:
-                expression = eval(parse_field(field))
+                expression = parse_expression(field)
                 self.expression_instance_cache[field] = expression
         except NameError as e:
             get_module_logger("data").exception(
--- a/qlib/data/expression_parser.py
+++ b/qlib/data/expression_parser.py
@@ -1,0 +1,136 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""Safe parser for qlib feature expression strings.
+
+The qlib feature DSL lets users write expressions such as::
+
+    Ref($close, -2) / Ref($close, -1) - 1
+
+which are transformed by :func:`qlib.utils.parse_field` into::
+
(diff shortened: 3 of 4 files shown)


andesyteoss:
Closing this as inactive — no maintainer response after 14 days.

The security finding and fix remain valid. If this is still relevant, I'm happy to reopen, rebase, or re-submit against a different branch. Just drop a comment.

PR #1736, opened 2024-01-20
Fix the bug of reading string NA as NaN in the function exists_qlib_data.

OzzyXu:
Fix the bug of reading string NA as NaN of exists_qlib_data in /qlib/utils/\_\_init\_\_.py.

## Description

Nano Labs Ltd is a new Nasdaq-listing company with the ticker name NA from August 1, 2022. The default na_value list of pd.read_csv includes "NA". Changed the default reading behavior of pd.read_csv in `exists_qlib_data` by adding `keep_default_na=False`. Removed two values ("NA" and "NULL") from the default NA list while reading the first column of "all.txt", which normally are all strings. 

## Motivation and Context

To fix the bug in #1720.

## How Has This Been Tested?

- [X] Pass the test by running: `pytest qlib/tests/test_all_pipeline.py` under upper directory of `qlib`.
- [X] If you are adding a new feature, test on your own test scripts.

## Screenshots of Test Results (if appropriate):
1. Pipeline test:
image

2. Your own tests:
image (1)
Place the attached file [all.txt](https://github.com/microsoft/qlib/files/13997190/all.txt) under \qlib_data\us_data_made\instruments and test with the following code 
```
import qlib
import pandas as pd
import sys, site
from pathlib import Path
from qlib.utils import exists_qlib_data
from qlib.constant import REG_US

scripts_dir = Path.cwd().parent.joinpath("scripts")
provider_uri = "./qlib_data/us_data_made"  # target_dir
if not exists_qlib_data(provider_uri):
    print(f"Qlib data is not found in {provider_uri}")
    sys.path.append(str(scripts_dir))
    from get_data import GetData

    GetData().qlib_data(target_dir=provider_uri, region=REG_US)
qlib.init(provider_uri=provider_uri, region=REG_US)
```

## Types of changes

- [X] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

PR #2302, opened 2026-07-18
fix(data): make FileInstrumentStorage round-trip instrument files correctly

### LINKED ISSUE #1720: Ticker name "NaN" sometimes makes the exists_qlib_data function report errors.
## 🐛 Bug Description

Occasionally the ticker name "NaN" in the "all.txt" under `/instruments` makes the `exists_qlib_data` function fail due to the string "NaN" being wrongly converted to the float "nan" but not a string. 

## To Reproduce

1. Save the attached [all.txt](https://github.com/microsoft/qlib/files/13803664/all.txt) under the `~/.qlib/qlib_data/us_data/instruments`.
2. Run the following code:
```python
provider_uri = "~/.qlib/qlib_data/us_data_new"  # target_dir
if not exists_qlib_data(provider_uri):
    print(f"Qlib data is not found in {provider_uri}")
    sys.path.append(str(scripts_dir))
    from get_data import GetData

    GetData().qlib_data(target_dir=provider_uri, region=REG_US)
``` 

## Expected Behavior

The code should run without errors. 

## Screenshot
image

## Environment

and paste them here directly.

 - Qlib version: 0.93
 - Python version: 3.8.10
 - OS (`Windows`, `Linux`, `MacOS`): Windows

## Additional Notes

1. For unknown reasons, sometimes the "NaN" string can be read correctly, e.g. with this [all.txt](https://github.com/microsoft/qlib/files/13803665/all.txt).
2. The bug is caused by the wrong usage of pandas.read_csv in the following line of `exists_qlib_data` under `qlib\utils\__init__.py`. Refer to the [page](https://github.com/pandas-dev/pandas/issues/20877) for more details. 
```python
 miss_code = set(pd.read_csv(_instrument, sep="\t", header=None).loc[:, 0].apply(str.lower)) - set(code_names)
``` 
3. The cause of the bug can be further verified by the following code:
```python
temp = pd.read_csv("all.txt", sep="\t", header=None).loc[:, 0]
non_string_values = [i for i in temp if not isinstance(i, str)]
print(non_string_values)
``` 
```
[nan]
``` 
4. The bug can be easily fixed by adding `keep_default_na=False`
```python
temp = pd.read_csv("all.txt", sep="\t", header=None, keep_default_na=False).loc[:, 0]
non_string_values = [i for i in temp if not isinstance(i, str)]
print(non_string_values)
``` 
```
[]
``` 
5. I can help with the fix, just want to ask what tests I need to run to make sure whether the fix would cause any other issues.

vignesh-nagarajan-vn:
Two long-standing defects in `FileInstrumentStorage` broke the instrument file round-trip:

- `_read_instrument` let pandas NA-parse the symbol column, so real tickers literally named "NA" or "NULL" were loaded as float NaN and merged into a single bogus instrument, which then leaked out of `D.list_instruments`. PR #1736 fixed this same defect in `exists_qlib_data` (reported in #1720), but the storage reader - the path that actually feeds data loading - kept the old behavior.

- `_write_instrument` wrote the file twice back-to-back; the second `to_csv` immediately overwrote the correctly-ordered output with columns in `[start, end, symbol]` order, which `_read_instrument` then misparsed (dates became symbols). Present since the original implementation (70fc5810); every public mutation (`update`, `__setitem__`, `__delitem__`) went through it.

**Fix:** disable default NA parsing when reading the symbol column (symbols are opaque identifiers), and write the file once in the documented `[symbol, start, end]` order, keeping the utf-8 encoding of the removed call.

Both defects are covered by new self-contained tests in `tests/storage_tests/test_storage.py` (temporary provider, no market data), each verified to fail against the previous code.

Behavior for well-formed data is unchanged: the existing `tests/storage_tests` pass as-is, `D.list_instruments` output on `qlib_data_simple` is hash-identical before/after the change, and black/flake8/pylint/mypy all pass with the CI invocations.

Fixes #1720 (completes the fix that #1736 applied only to `exists_qlib_data`)

### DIFF
--- a/qlib/data/storage/file_storage.py
+++ b/qlib/data/storage/file_storage.py
@@ -212,6 +212,9 @@ class FileInstrumentStorage(FileStorageMixin, InstrumentStorage):
             names=[self.SYMBOL_FIELD_NAME, self.INSTRUMENT_START_FIELD, self.INSTRUMENT_END_FIELD],
             dtype={self.SYMBOL_FIELD_NAME: str},
             parse_dates=[self.INSTRUMENT_START_FIELD, self.INSTRUMENT_END_FIELD],
+            # Symbols are opaque identifiers: tickers literally named "NA" or "NULL" must not be
+            # read as missing values (#1736 fixed the same defect in `exists_qlib_data` only).
+            keep_default_na=False,
         )
         for row in df.itertuples(index=False):
             _instruments.setdefault(row[0], []).append((row[1], row[2]))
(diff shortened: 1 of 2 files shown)


vignesh-nagarajan-vn:

PR #2312, opened 2026-08-01
Add CausalWM357: a weight-free, discovered feature handler (357 features)

kayuksel:
# Add `CausalWM357`: 357 weight-free, closed-form, discovered features

## What this adds

| File | Purpose |
|---|---|
| `qlib/contrib/data/causalwm_handler.py` | `CausalWM357(DataHandlerLP)` + `CausalWM357DL(DataLoader)` |
| `tests/data_mid_layer_tests/test_causalwm_handler.py` | optional-dependency guard, data-store construction, frame contract, no-NaN-explosion, causal-lag property |
| `examples/benchmarks/LightGBM/workflow_config_lightgbm_CausalWM357.yaml` | runnable benchmark config, structurally identical to the Alpha158 one |

`CausalWM357` is an `Alpha158`/`Alpha360` sibling — same constructor shape, same
`feature`/`label` contract — built on a feature set that was **discovered, not
hand-engineered**, and that carries **zero learned parameters**.

The name follows the existing convention of naming a handler after its feature
count: 111 closed-form programs producing 357 columns.

## Why

Qlib's built-in feature libraries are hand-written expression bags. This one was
produced by an *explanatory* search: a candidate program was admitted only if it
improved cross-validated prediction of the market's own **next state**, never by
fitting returns. What survives is a small set of frozen closed forms with three
properties that matter for a library feature set:

1. **Weight-free and reproducible.** Nothing is trained at load time. The
   handler re-executes fixed formulas; the same input always yields the same
   output, on CPU, with no checkpoint to ship or version.
2. **Auditable end to end.** Every column is named after the program that
   produced it — a feature-importance table reads as mechanisms, not as
   `KLEN`/`VSTD5` codes. `python -m causalwm --show ret_weighted_reversal`
   prints the exact source of any column's program.
3. **Slow, transferable structure.** The programs were discovered on a US
   universe and applied unchanged to CSI 300; the measurements below are all of
   that transferred feature set, with no re-discovery on the target market.

## How it differs from Alpha158 / Alpha360

| | Alpha158 / Alpha360 | CausalWM357 |
|---|---|---|
| Definition | expressions in the Qlib operator DSL | 111 Python closed-form programs (in the `causalwm` package) |
| Computation | `QlibDataLoader` | small custom `DataLoader` (the ops have no DSL form) |
| Row grid | daily | **weekly, W-FRI, Friday-stamped** (daily bars resampled internally) |
| Scope | per-instrument | **cross-sectional** over the loaded universe (per-week Gaussianized ranks) |
| Universe | any | **full-history only** — instruments entering/leaving mid-span are dropped, with a warning |
| Horizon | daily label | weekly label; the feature set was discovered at a **two-week** horizon |
| Dependencies | none beyond qlib | optional `causalwm` (brings torch, CPU is fine) |

Two design points worth reviewing explicitly:

* **Causal availability lag.** The phi row computed from weekly bars through
  Friday *W* is stamped at Friday `W + availability_lag_weeks`, default **1**:
  the row served at Friday *W* was already fully computable a week earlier, so
  it needs no same-bar data to be tradable at *W*'s close. `0` restores the
  source paper's protocol. Either way `feature[t]` depends only on raw data at
  or before *t*; there is a test for this that removes future data and asserts
  no earlier row changes.
* **Label.** The default is Alpha158's own convention,
  `Ref($close, -2)/Ref($close, -1) - 1`, read **on the weekly grid**. Any
  `Ref($close, -a)/Ref($close, -b) - 1` is accepted (so the paper's two-week
  horizon is `Ref($close, -2)/$close - 1`). Because the loader builds weekly
  bars in Python it does not run Qlib's expression engine, so any other
  expression is rejected with an explanatory `NotImplementedError` rather than
  silently mis-evaluated.

## The optional-dependency story

The feature math lives in a separate MIT-licensed pip package, `causalwm`
(PyTorch is used only as an array library for a weight-free forward pass).
**Qlib gains no new hard dependency and no new numerics to maintain** — the file
added here is a thin adapter.

* the module imports cleanly with neither `causalwm` nor `torch` installed;
* both are imported lazily, only when features (or feature names) are actually
  requested;
* the failure mode is an `ImportError` that names the fix:

```
CausalWM357 needs the optional `causalwm` package (which brings PyTorch; CPU is fine):
    pip install causalwm
It ships the 111 closed-form feature programs; they carry ZERO learned weights, torch is used only as an array library for the forward pass.
```

For precedent: 33 modules under `qlib/contrib/` already import torch in
0.9.7 — we still kept the import lazy, because a data handler should not be a
reason to install a deep-learning framework.

## How to run

```bash
pip install causalwm
qrun examples/benchmarks/LightGBM/workflow_config_lightgbm_CausalWM357.yaml
```

or in Python:

```python
from qlib.contrib.data.causalwm_handler import CausalWM357

h = CausalWM357(
    instruments="csi300",
    start_time="2017-01-01", end_time="2020-08-01",
    fit_start_time="2017-01-01", fit_end_time="2018-12-31",
)
df = h.fetch()          # (weekly rows) x (357 features + LABEL0)
```

Tests:

```bash
pytest tests/data_mid_layer_tests/test_causalwm_handler.py
```

Without `causalwm` installed, the data-store tests skip and the
optional-dependency tests still run and pass.

## Numbers we measured

These are **our** measurements, stated with their protocol. They are *not*
produced by the Qlib workflow above; they come from the harness of the paper
this feature set is from, which is stricter than a single held-out split.

**Protocol.** CSI 300, a fixed 225-name full-history universe, weekly bars,
two-week forward log-return Gaussianized cross-sectionally, 28-path
combinatorially-purged CPCV (8 blocks choose 2, 3-week embargo on both sides),
ridge readout as the only fitted object, refit per market. Every feature set
saw the **same universe, same folds, same readout**; Alpha158/Alpha360 were
extracted from Qlib and evaluated at each set's own best in-fold configuration.

| Feature set | mean rank IC (28 paths) | worst path | paths > 0 |
|---|---|---|---|
| Alpha360 (300 cols) | +0.014 | −0.011 | 24/28 |
| Alpha158 (157 cols) | +0.045 | +0.021 | 28/28 |
| **CausalWM357 (357 cols, transferred)** | +0.0385 | −0.015 | 27/28 |
| **CausalWM357 + in-fold top-40 selection** | **+0.047** | **+0.001** | **28/28** |

Selection is applied inside each training fold (no leakage). It *raises* the
discovered core (+0.0385 → +0.047) and *lowers* Alpha158 (+0.045 → +0.037 at
top-40), which is the behaviour you would expect from a separable set of
mechanisms versus a homogeneous engineered bag.

**State self-prediction.** Same weekly panels, one harness, CSI 300: fit an
affine ridge transition operator on each feature state and score out-of-sample
self-prediction, plus the half-life of an autoregressive rollout from a single
anchor (no observations after the anchor).

| State | R² @ 2 wk | R² @ 8 wk | R² @ 26 wk | rollout half-life |
|---|---|---|---|---|
| Alpha158 | 0.26 | 0.05 | 0.01 | 3.8 wk |
| Alpha360 | 0.48 | 0.07 | 0.00 | 4.8 wk |
| **CausalWM357** | **0.77** | **0.50** | **0.21** | **20.2 wk** |

This is the property the search optimized for, and it is the honest reason to
consider the set interesting as a *state* rather than as another alpha bag.

### What this feature set is not

* **It is a two-week signal, not a daily alpha.** At the one-day horizon it
  carries very little: rank IC ≈ 0.007–0.025 depending on protocol. In Qlib's
  own daily pipeline on the public `cn_data` store, under a linear readout, we
  measured 1-day rank IC +0.0085 for this set against −0.0007 (Alpha158) and
  −0.0089 (Alpha360) — i.e. everything is near zero at that horizon. Anyone
  benchmarking daily strategies should expect little from it.
* **The benchmark yaml has not been tuned.** It mirrors the Alpha158 LightGBM
  config so the comparison is structural, not a hyper-parameter contest; the
  only deliberate changes are `ann_scaler: 52` (weekly rows) and the fit-ranged
  `RobustZScoreNorm`.
* **The universe restriction is real.** Requiring a gapless 52-week history per
  instrument drops names that enter or leave mid-span. Per-window dynamic
  universe support is future work, and we are happy to do it in this PR if
  maintainers prefer.
* **We have not re-run the paper's CPCV numbers inside Qlib's workflow.** The
  table above is our harness; the Qlib-pipeline comparison mentioned in the
  1-day bullet is a separate, linear-readout run.

## Alternative, if you prefer

If maintainers would rather not carry a handler with an optional dependency,
the same class ships in the `causalwm` package itself and works today through
`module_path: causalwm.qlib_handler` with no change to qlib. We are proposing it
here because a discovered, weight-free, auditable feature set seems like a
useful third point next to Alpha158 and Alpha360 — but the package route is a
perfectly good outcome and we will close the PR without argument.

## Verification (run 2026-08-01, macOS arm64, Python 3.10, qlib 0.9.7 + torch 2.13.0)

Against a real Qlib CSI 300 store (`~/.qlib/qlib_data/cn_data`):

| Check | Result |
|---|---|
| Handler test-suite (`test_causalwm_handler.py`) | **9 passed, 1 skipped** (the skip is the torch-absent case, unreachable when torch is installed) |
| Column contract vs the reference implementation | `CausalWM357` matches **Alpha158** exactly: `_data` is 2-level, `fetch()` returns 1 level (Qlib's `CS_ALL` drops the outer level), `fetch(col_set=CS_RAW)` is 2-level |
| `qrun workflow_config_lightgbm_CausalWM357.yaml` | **completes end to end**: IC 0.0709, ICIR 0.4732, Rank IC 0.0574, Rank ICIR 0.3273 |
| Causal-lag property (features at *t* use only data <= *t*) | max abs. difference **0.000e+00** over 4,720 shared rows |
| Encoder vs the discovery engine's reference features | **max abs. difference 6.1e-05** over 220,428 windows x 472 features |

The `qrun` figures above are a **smoke test, not a benchmark**: one 2020 test split on a
locally truncated store (it ends 2020-09), Qlib's daily label, LightGBM. They demonstrate
that the workflow runs and produces sane output. The evaluation we stand behind is the
purged-CPCV comparison in the section above.

### DIFF
--- a/examples/benchmarks/LightGBM/workflow_config_lightgbm_CausalWM357.yaml
+++ b/examples/benchmarks/LightGBM/workflow_config_lightgbm_CausalWM357.yaml
@@ -1,0 +1,96 @@
+# LightGBM on CausalWM357, in the exact layout of
+# examples/benchmarks/LightGBM/workflow_config_lightgbm_Alpha158.yaml.
+#
+#   pip install causalwm      # optional dependency: the 111 feature programs
+#   qrun examples/benchmarks/LightGBM/workflow_config_lightgbm_CausalWM357.yaml
+#
+# Differences from the Alpha158 config, all forced by the handler and all
+# deliberate (see qlib/contrib/data/causalwm_handler.py):
+#   * rows are WEEKLY (W-FRI, Friday-stamped): the model trains and predicts
+#     once a week, so the backtest rebalances weekly, and SigAnaRecord's
+#     ann_scaler is 52 (weeks/year) instead of 252 (days/year);
+#   * the label is Alpha158's Ref($close, -2)/Ref($close, -1) - 1 read on the
+#     weekly grid (buy next week's close, sell the week after);
+#   * features are served RAW, so a fit-ranged RobustZScoreNorm is required
+#     (this is the handler default, spelled out here for visibility);
+#   * the handler over-fetches 420 calendar days before start_time for the
+#     encoder's 52-week window, so the segments below need no extra warm-up.
+qlib_init:
+    provider_uri: "~/.qlib/qlib_data/cn_data"
+    region: cn
+market: &market csi300
+benchmark: &benchmark SH000300
+data_handler_config: &data_handler_config
+    start_time: 2008-01-01
+    end_time: 2020-08-01
+    fit_start_time: 2008-01-01
+    fit_end_time: 2014-12-31
+    instruments: *market
+    infer_processors:
+        - class: RobustZScoreNorm
+          kwargs:
+              fields_group: feature
+              clip_outlier: true
+        - class: Fillna
+          kwargs:
+              fields_group: feature
+port_analysis_config: &port_analysis_config
+    strategy:
+        class: TopkDropoutStrategy
+        module_path: qlib.contrib.strategy
+        kwargs:
+            signal: <PRED>
+            topk: 50
+            n_drop: 5
+    backtest:
+        start_time: 2017-01-01
+        end_time: 2020-08-01
+        account: 100000000
+        benchmark: *benchmark
+        exchange_kwargs:
+            limit_threshold: 0.095
+            deal_price: close
+            open_cost: 0.0005
+            close_cost: 0.0015
+            min_cost: 5
+task:
+    model:
+        class: LGBModel
+        module_path: qlib.contrib.model.gbdt
+        kwargs:
+            loss: mse
+            colsample_bytree: 0.8879
+            learning_rate: 0.2
+            subsample: 0.8789
+            lambda_l1: 205.6999
+            lambda_l2: 580.9768
+            max_depth: 8
+            num_leaves: 210
+            num_threads: 20
+    dataset:
+        class: DatasetH
+        module_path: qlib.data.dataset
+        kwargs:
+            handler:
+                class: CausalWM357
+                module_path: qlib.contrib.data.causalwm_handler
+                kwargs: *data_handler_config
+            segments:
+                train: [2008-01-01, 2014-12-31]
+                valid: [2015-01-01, 2016-12-31]
+                test: [2017-01-01, 2020-08-01]
+    record:
+        - class: SignalRecord
+          module_path: qlib.workflow.record_temp
+          kwargs:
+            model: <MODEL>
+            dataset: <DATASET>
+        - class: SigAnaRecord
+          module_path: qlib.workflow.record_temp
+          kwargs:
+            ana_long_short: False
+            ann_scaler: 52
+        - class: PortAnaRecord
+          module_path: qlib.workflow.record_temp
+          kwargs:
+            config: *port_analysis_config
--- a/qlib/contrib/data/causalwm_handler.py
+++ b/qlib/contrib/data/causalwm_handler.py
@@ -1,0 +1,249 @@
+# Copyright (c) Microsoft Corporation.
+# Licensed under the MIT License.
+"""CausalWM357 — 357 weight-free, closed-form world-model features.
+
+``CausalWM357`` is an ``Alpha158``/``Alpha360`` sibling built on a *discovered*
+feature set: 111 closed-form programs producing 357 columns, found by an
+explanatory search (does the candidate improve prediction of the market's own
+next state?) rather than by fitting returns.  **There are no learned weights
+anywhere** — every column is a frozen closed form, so the handler re-executes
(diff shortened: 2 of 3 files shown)


kayuksel:

kayuksel:


Kamer Ali Yuksel
Head of AI Lab <https://www.linkedin.com/in/kyuksel/>
Timezone: GMT+2

kayuksel:
Here is the publication related with this work, if you'd like to know more details: https://ssrn.com/abstract=7216139

PR #2313, opened 2026-08-04
fix(processor): use fields_group in TanhProcess to protect labels

### LINKED ISSUE #1687: class TanhProcess(Processor) error

ali2002-collab:
## Description
`TanhProcess` previously selected columns with `get_level_values(1).str.contains("LABEL")`, which is inconsistent with Qlib's handler MultiIndex layout (`feature` / `label` on level 0, column names like `LABEL0` on level 1) and with other processors.

This change makes `TanhProcess` take `fields_group` (default `"feature"`) and transform only that group via `get_group_columns`, applying `tanh(x - 1)` to features while leaving labels untouched. `fields_group=None` still transforms all columns.

## Motivation and Context
Fixes #1687.

The old LABEL substring mask is brittle. Simply switching to level 0 would also be wrong for standard handlers, because level 0 uses lowercase `"label"`, not `"LABEL"`. Using `fields_group` matches the pattern used by `Fillna`, `CSZScoreNorm`, and related processors.

## How Has This Been Tested?

Ran:
`pytest tests/data_mid_layer_tests/test_processor.py::TestTanhProcess -v`

Also checked Black (`-l 120`), flake8, and pylint on the touched processor module.

1. Pipeline test:
N/A (processor unit tests; no market data required)

2. Your own tests:
Screenshot 2026-08-05 at 00 37 01

## Types of changes
- [x] Fix bugs
- [ ] Add new feature
- [x] Update documentation

### DIFF
--- a/docs/component/data.rst
+++ b/docs/component/data.rst
@@ -433,7 +433,7 @@ The ``Processor`` module in ``Qlib`` is designed to be learnable and it is respo
 
 - ``DropnaProcessor``: `processor` that drops N/A features.
 - ``DropnaLabel``: `processor` that drops N/A labels.
-- ``TanhProcess``: `processor` that uses `tanh` to process noise data.
+- ``TanhProcess``: `processor` that applies ``tanh(x - 1)`` to denoise values. By default it transforms the ``feature`` column group only (via ``fields_group``), leaving labels unchanged.
 - ``ProcessInf``: `processor` that handles infinity values, it will be replaces by the mean of the column.
 - ``Fillna``: `processor` that handles N/A values, which will fill the N/A value by 0 or other given number.
 - ``MinMaxNorm``: `processor` that applies min-max normalization.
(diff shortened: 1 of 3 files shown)


ali2002-collab:

Issue #2315, opened 2026-08-11
Benchmark weighting: default backtest compares an equal-weighted portfolio against a cap-weighted index


taro0915:
## 1. Summary

  In the default backtest configuration, `TopkDropoutStrategy` allocates cash equally across
  holdings while the reported excess return subtracts a cap-weighted index (`SH000300`) as a plain
  difference. Over 2017-01-03..2020-07-31 that weighting mismatch alone is worth **-4.13% per year**
  (information ratio -0.723) with no alpha involved, measured by simply holding the point-in-time
  CSI300 constituents equally weighted. Because equal weighting underperformed the cap-weighted
  index in this window, the offset makes reported model excess returns *lower* than a
  weighting-consistent comparison would show, so this is a measurement-convention question rather
  than a result that flatters Qlib.

  ## 2. Evidence in code

  Repository commit `79633dd9506ea689e5400dea0197717b5b3d74b7` (2026-07-23).

  * **Portfolio is equal-weighted by cash.** `qlib/contrib/strategy/signal_strategy.py:266`

        value = cash * self.risk_degree / len(buy) if len(buy) > 0 else 0

  * **Excess return is a plain difference against the benchmark series.**
    `qlib/workflow/record_temp.py:507-512`

        analysis["excess_return_without_cost"] = risk_analysis(
            report_normal["return"] - report_normal["bench"], freq=_analysis_freq
        )
        analysis["excess_return_with_cost"] = risk_analysis(
            report_normal["return"] - report_normal["bench"] - report_normal["cost"], freq=_analysis_freq
        )

  * **The shipped example configs set a cap-weighted index as that benchmark.**
    `examples/benchmarks/LightGBM/workflow_config_lightgbm_Alpha158.yaml:5` (`benchmark: &benchmark SH000300`)
    and line 23 inside `backtest:`. The same default appears in `qlib/workflow/record_temp.py:410`.

  * **An equal-weight benchmark is anticipated but not implemented.**
    `qlib/backtest/profit_attribution.py:238`

        :param bench: The benchmark for comparing. TODO: if no benchmark is set, the equal-weighted is used.

  I searched the GitHub issue tracker via the search API for existing discussion of benchmark
  weighting and did not find a matching issue; apologies if I missed one.

  ## 3. Measurement

  ### 3a. Model-independent control

  To isolate the weighting convention from any alpha, I held the point-in-time CSI300 constituents
  equally weighted and compared them with `SH000300` over the same window the example configs
[...]

PR #2316, opened 2026-08-11
Docs: clarify equal-weighted portfolio vs cap-weighted benchmark comp…

taro0915:
## Description

Documentation-only follow-up to #2315. No behavior change.

- `qlib/contrib/strategy/signal_strategy.py` — add a class docstring to `TopkDropoutStrategy` stating that positions are equal-weighted, and noting the implication for benchmark choice. (The class previously had no docstring, only a `# TODO:` block.)
- `qlib/workflow/record_temp.py` — add a note to the `PortAnaRecord` docstring describing how `excess_return_with_cost` / `excess_return_without_cost` are computed, and that no weighting adjustment is applied.
- `docs/component/strategy.rst` — add the same note to the `TopkDropoutStrategy` section.

No code paths, defaults, or outputs are modified.

## Motivation and Context

Related issue: #2315

`TopkDropoutStrategy` sizes new positions by splitting the available cash evenly across the stocks to be bought (`qlib/contrib/strategy/signal_strategy.py:266`), so the realized portfolio is close to equal-weighted. `PortAnaRecord` reports excess return as a plain arithmetic difference against a single benchmark instrument (`qlib/workflow/record_temp.py:507-512`), and the default benchmark is `SH000300` (CSI 300), which is capitalization-weighted.

These two defaults are used together in the default `PortAnaRecord` config and throughout `examples/benchmarks`. Because no adjustment is made for the difference in weighting scheme, the reported excess return contains the equal- versus cap-weighting spread of the universe in addition to the contribution of the signal. #2315 has the measurements: with a zero-alpha signal the offset is material and consistently signed, and it shifts the reported numbers for a real LightGBM + Alpha158 run as well.

This is not a defect in the arithmetic. Excess return against a named index is a legitimate quantity, and it is the right one if the question is "did this beat the CSI 300." It is a gap in what the documentation lets a reader assume, because the same number is routinely read as the contribution of the model. This PR makes the distinction explicit at the three places a user is most likely to be standing when they form that assumption.

### Why documentation first

#2315 sketched two possible directions: changing or parameterizing the comparison, or documenting the current behavior. The second is separable and carries no compatibility risk, so it is offered on its own here rather than bundled. If maintainers prefer the first, an equal-weighted benchmark option is a natural follow-up — `qlib/contrib/report/analysis_position/profit_attribution.py:238` already carries a TODO for exactly that — and I am happy to open it.

## How Has This Been Tested?

Docstrings and reStructuredText only; no executable code paths are touched, so there are no test coverage implications. The change adds lines only (+40 / -0) and does not modify any existing line, so `black` formatting of existing code is unaffected.

## Types of changes

- [ ] Fix bugs
- [ ] Add new feature
- [x] Update documentation

### DIFF
--- a/docs/component/strategy.rst
+++ b/docs/component/strategy.rst
@@ -88,6 +88,14 @@ TopkDropoutStrategy
 
 - Generate the order list from the target amount
 
+.. note::
+    `TopkDropoutStrategy` holds an **equal-weighted** portfolio: the available cash is split
+    evenly among the stocks bought on each rebalance. When its results are compared against a
+    capitalization-weighted benchmark such as the default ``SH000300``, the reported excess
+    return also contains the equal- versus cap-weighting spread of the universe, which is a
+    property of the weighting scheme rather than of the signal. Use a benchmark whose weighting
+    scheme matches the strategy if you want to isolate the contribution of the signal.
+
 EnhancedIndexingStrategy
 ------------------------
 `EnhancedIndexingStrategy` Enhanced indexing combines the arts of active management and passive management,
--- a/qlib/contrib/strategy/signal_strategy.py
+++ b/qlib/contrib/strategy/signal_strategy.py
@@ -73,6 +73,24 @@ class BaseSignalStrategy(BaseStrategy, ABC):
 
 
 class TopkDropoutStrategy(BaseSignalStrategy):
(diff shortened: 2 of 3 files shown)


taro0915:

PR #2322, opened 2026-08-17
Fix daily batching for time-series samplers

### LINKED ISSUE #2319: DailyBatchSampler in pytorch_gats_ts.py yields cross-day batches: TSDataSampler.get_index() swaps index labels but not row order
## Bug description

`DailyBatchSampler` (`qlib/contrib/model/pytorch_gats_ts.py`) assumes rows belonging to the same trading day are **contiguous** in the data source: it computes per-day counts via `groupby("datetime").size()`, converts them to start offsets with `cumsum`, and slices contiguous ranges `np.arange(idx, idx + count)`.

However, `TSDataSampler` stores data instrument-major — `qlib/data/dataset/__init__.py` builds it as `self.data = data.swaplevel().sort_index()` (and the `data_index` docstring itself says `index order <instrument, datetime>`). `get_index()` then returns `self.data_index.swaplevel()`, which swaps the **labels** of the MultiIndex back to `(datetime, instrument)` but does **not** reorder rows.

As a result each "daily" batch actually contains **one instrument across many consecutive days**, not one day's cross-section. Any model trained with this sampler (e.g. the GATs benchmark, and code copying this sampler) silently trains its graph/attention over "one stock's history" instead of "one day's cross-section".

## Minimal reproduction

```python
import numpy as np
import pandas as pd

dates = pd.date_range("2020-01-01", periods=5, freq="B")
insts = ["A", "B", "C"]
idx = pd.MultiIndex.from_product([dates, insts], names=["datetime", "instrument"])
df = pd.DataFrame({"f": range(len(idx)), "label": range(len(idx))}, index=idx)

s = TSDataSampler(df, dates[0], dates[-1], step_len=2)
print(s.get_index()[:6].tolist())
# [(2020-01-01, A), (2020-01-02, A), (2020-01-03, A), (2020-01-06, A), (2020-01-07, A), (2020-01-01, B)]
# -> labels say (datetime, instrument), but rows are instrument-major.

# DailyBatchSampler's first "day" batch is rows 0..2 (3 instruments expected),
# which are actually instrument A on three different days.
```

Note the docstring of `get_index()` explicitly advertises the day-by-day use case: *"Special sampler will be used (e.g. user want to sample day by day)"*.

## Expected behavior

Either:
1. `get_index()` documents that row order is instrument-major (and `DailyBatchSampler` is fixed to not assume contiguity), or
2. `DailyBatchSampler` groups actual row positions by datetime instead of assuming contiguous blocks.

## Suggested fix (drop-in for DailyBatchSampler)

```python
class DailyBatchSampler(Sampler):
    def __init__(self, data_source):
        self.data_source = data_source
        index = data_source.get_index()
        positions = pd.Series(np.arange(len(index)), index=index.get_level_values("datetime"))
        self.batches = [g.to_numpy() for _, g in positions.groupby(level=0, sort=True)]

    def __iter__(self):
        yield from self.batches

    def __len__(self):
        return len(self.batches)
```

When using this, downstream index alignment for predictions must follow the sampler's iteration order, e.g. `dl_test.get_index()[np.concatenate(self.batches)]`.

## Environment

Verified against current `main`: `qlib/data/dataset/__init__.py` (`get_index`, and `data.swaplevel().sort_index()` in `TSDataSampler.__init__`) and `qlib/contrib/model/pytorch_gats_ts.py` (`DailyBatchSampler`) are unchanged. Found while implementing a cross-sectional model whose per-day-batch unit test ("each batch contains exactly one datetime") failed against the GATs-style sampler.

sergio12S:
`TSDataSampler.get_index()` exposes datetime/instrument labels over an
instrument-major row layout. `DailyBatchSampler` previously calculated correct
per-date counts but converted them to contiguous ranges, so a nominal daily
batch could mix multiple dates.

I reproduced this with five dates and three instruments: every emitted batch
contained three distinct dates. This change groups the actual source-row
positions by datetime. Because that changes iteration order, `predict()` now
builds its result index from the sampler's order as well, preventing otherwise
correct predictions from being assigned to the wrong observations. `__len__`
now returns the number of daily batches.

The added regression test constructs a real `TSDataSampler` and checks batch
membership, sampler length, and output-index alignment.

Tested with:

`python -m pytest -q tests/model/test_pytorch_gats_ts.py` — 1 passed.

Closes #2319.

### DIFF
--- a/qlib/contrib/model/pytorch_gats_ts.py
+++ b/qlib/contrib/model/pytorch_gats_ts.py
@@ -26,19 +26,20 @@ from ...contrib.model.pytorch_gru import GRUModel
 class DailyBatchSampler(Sampler):
     def __init__(self, data_source):
         self.data_source = data_source
-        # calculate number of samples in each batch
-        self.daily_count = (
-            pd.Series(index=self.data_source.get_index()).groupby("datetime", group_keys=False).size().values
-        )
-        self.daily_index = np.roll(np.cumsum(self.daily_count), 1)  # calculate begin index of each batch
-        self.daily_index[0] = 0
+        index = self.data_source.get_index()
+        positions = pd.Series(np.arange(len(index)), index=index)
+        self.daily_index = [group.values for _, group in positions.groupby(level="datetime", sort=True)]
 
     def __iter__(self):
-        for idx, count in zip(self.daily_index, self.daily_count):
-            yield np.arange(idx, idx + count)
+        yield from self.daily_index
 
     def __len__(self):
-        return len(self.data_source)
+        return len(self.daily_index)
+
+    def get_index(self):
+        """Return the data source index in sampler iteration order."""
+        positions = np.concatenate(self.daily_index)
+        return self.data_source.get_index()[positions]
 
 
 class GATs(Model):
@@ -332,7 +333,7 @@ class GATs(Model):
 
             preds.append(pred)
 
-        return pd.Series(np.concatenate(preds), index=dl_test.get_index())
+        return pd.Series(np.concatenate(preds), index=sampler_test.get_index())
 
 
 class GATModel(nn.Module):
--- a/tests/model/test_pytorch_gats_ts.py
+++ b/tests/model/test_pytorch_gats_ts.py
@@ -1,0 +1,25 @@
+import pandas as pd
+
+from qlib.contrib.model.pytorch_gats_ts import DailyBatchSampler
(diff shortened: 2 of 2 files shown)


koriyoshi2041 (approved):
Verified at 1d58db79. Grouping source-row positions by datetime fixes the instrument-major layout bug, and using the same sampler order for the prediction index keeps values aligned. The regression covers daily membership, batch count, and index ordering; the focused test passes locally, and the diff check is clean.

PR #2327, opened 2026-08-19
Fix CatBoost and XGBoost evals_result handling

mameikagou:
## Description

Fix evaluation-result handling in the CatBoost and XGBoost model wrappers.

- `CatBoostModel.fit` now preserves and updates a caller-provided `evals_result` dictionary instead of rebinding the local variable and silently discarding the output.
- Both wrappers now use `None` instead of a mutable dictionary as the default argument.
- Add focused, network-free regression tests for caller mutation and fresh defaults.

## Root cause

`CatBoostModel.fit` assigned `self.model.get_evals_result()` to the local `evals_result` name. When a caller passed a dictionary, that dictionary was never populated. Both wrappers also declared `evals_result=dict()` in the function signature, sharing one dictionary across calls that omit the argument.

## Impact

Callers can reliably collect train and validation metric histories through the same `evals_result` contract used by the other Qlib model wrappers. Existing calls that omit the argument remain compatible.

## Validation

- Regression check against the previous implementation: 4 failures / 1 pass
- Patched focused tests: `5 passed`
- Black (`--target-version py38 --line-length 120`): passed
- Flake8 with the repository ignore set: passed

## Risk

Low. The model training calls and returned model objects are unchanged; the patch only corrects the optional output dictionary and removes mutable defaults.

### DIFF
--- a/qlib/contrib/model/catboost_model.py
+++ b/qlib/contrib/model/catboost_model.py
@@ -31,10 +31,12 @@ class CatBoostModel(Model, FeatureInt):
         num_boost_round=1000,
         early_stopping_rounds=50,
         verbose_eval=20,
-        evals_result=dict(),
+        evals_result=None,
         reweighter=None,
         **kwargs,
     ):
+        if evals_result is None:
+            evals_result = {}
         df_train, df_valid = dataset.prepare(
             ["train", "valid"],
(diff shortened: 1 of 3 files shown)


mameikagou:

PR #2339, opened 2026-09-02
fix(security)!: harden artifact serialization and loading

XianBW:
## Summary

- Replace HIST's pickled stock-index mapping with validated JSON and fix CPU execution.
- Use an explicit pickle allowlist for supported NumPy/pandas data, including record arrays and business-calendar frequencies, instead of trusting entire namespaces.
- Make recorder artifact loading restricted by default, with consistent `trusted` opt-ins for workflows and record templates, plus per-artifact collector consent. Automatic prediction/label reads remain restricted.
- Constrain `HighFreqProvider`'s configured and generated artifact paths to `artifact_root`.
- Correct DDG-DA cache replay, rolling tails, IC indexing, path handling and shared-template mutation; support disabled LightGBM early stopping and remove an unused PIT calendar dependency.
- Add security regressions and complete HIST/DDG-DA workflow coverage, plus explicit CI dataset downloads with bounded retries.

## Upgrade notes

**Breaking change:** Reloading saved models, datasets and tasks containing executable objects requires explicit `trusted=True` (`--trusted=True` in example CLIs), after verifying their source and storage. Previously saved workflow components without consent default to restricted loading.

HIST's `stock_index` must use JSON, and high-frequency artifact paths must remain within the configured root. The [migration guide](https://qlib.readthedocs.io/en/latest/start/artifact_migration.html) explains component-level authorization and data-format compatibility, with examples for restored managers, Arrow columns and timezones.

PR #2340, opened 2026-09-02
fix(security): constrain config-driven code execution

XianBW:
## Changes

- Replace `eval()` in feature expressions and disk expression caching with a restricted AST interpreter. Preserve registered operators and safe parameter syntax, including bounded containers, argument unpacking, indexing/slicing, and scalar conditions.
- Preserve quoted strings and necessary token boundaries during expression preprocessing and cache-key normalization.
- Replace TRA model and performance-graph name evaluation with explicit mappings.
- Require explicit, strictly boolean `trusted=True` for file-module imports, defaulting to refusal. Keep legacy module aliases for restoring trusted older file-based models.
- Update affected TRA/LightGBM examples, add regression coverage, and connect migration guidance to documentation entry points and actionable errors.

## Compatibility and migration

Package imports and standard Alpha158/Alpha360 expressions remain supported without opt-in. File-based components must explicitly authorize their imports:

```yaml
task:
  model:
    class: MyModel
    module_path: custom_modules/model.py
    trusted: true
    kwargs: {}
```

Direct imports use `get_module_by_module_path(path, trusted=True)`. Each nested file component needs its own consent; there is no global grant. Top-level `trusted` controls module loading, while `kwargs.trusted` remains a constructor argument.

Configurations and imported code must still be trusted. This is not a sandbox or directory-containment policy, and module-import consent does not enable unrestricted expressions or authorize artifact deserialization. Artifact-loading changes are handled separately in #2339.

See the [configuration execution migration guide](https://github.com/microsoft/qlib/blob/security/constrain-config-execution/docs/start/config_migration.rst) for supported syntax, custom extensions, and old-model recovery.

### DIFF
--- a/.github/workflows/lint_title.yml
+++ b/.github/workflows/lint_title.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v4
         with:
-          node-version: '16'
+          node-version: '22'
 
       - name: Install commitlint
         run: npm install --save-dev @commitlint/{config-conventional,cli}
--- a/.github/workflows/test_qlib_from_pip.yml
+++ b/.github/workflows/test_qlib_from_pip.yml
@@ -13,6 +13,9 @@ on:
 jobs:
   build:
     timeout-minutes: 120
+    env:
+      # Released Qlib versions use MLflow's legacy file-store backend.
+      MLFLOW_ALLOW_FILE_STORE: "true"
 
     runs-on: ${{ matrix.os }}
     strategy:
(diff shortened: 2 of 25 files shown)


XianBW:
Local verification update:

- 18 security regression tests passed
- security tests plus handler serialization and full train/backtest pipeline: 22 passed
- repository-wide Black, Flake8, and Pylint passed
- all eight modified example YAML files parsed successfully
- removed the optional Plotly import from the registry regression test so test collection is independent of the analysis extra

The branch also contains minimal CI compatibility fixes for current Node/commitlint, Plotly, MLflow, and Windows/Python 3.8 OSQP behavior.

XianBW:
Final verification: all 76 GitHub checks pass on commit `ee6b4518`/latest branch state, including title, CLA, pip compatibility, source builds, notebook execution, full pytest, and slow pytest matrices.

PR #2345, opened 2026-09-10
docs: add alternative community data source

xu-duqing:
## Summary

- add `xu-duqing/investment_data` to the community data sources in Data Preparation
- clarify that the existing download commands work with either repository

## Verification

- verified both `releases/latest/download/qlib_bin.tar.gz` URLs return HTTP 200
- ran `git diff --check`


xu-duqing:

PR #2348, opened 2026-09-11
fix(backtest): normalize freq before querying benchmark feature data

### LINKED ISSUE #1925: backtest occur errror while executor's time_per_step=1day
## To Reproduce

```python


__benchmark = "SH000300"

portfolio_analysis_config = {
    "strategy": {
        "class": "TopkDropoutStrategy",
        "module_path": "qlib.contrib.strategy",
        "kwargs": {
            "signal": '<PRED>',
            "topk": 50,
            "n_drop": 5,
        },
    },
    "executor": {
        "class": "SimulatorExecutor",
        "module_path": "qlib.backtest.executor",
        "kwargs": {
            "time_per_step": "1day",
            "generate_portfolio_metrics": True,
        },
    },
    "backtest": {
        "start_time": "2017-01-01",
        "end_time": "2020-08-01",
        "account": 100000000,
        "benchmark": __benchmark,
        "exchange_kwargs": {
            "limit_threshold": 0.095,
            "deal_price": "close",
            "open_cost": 0.0005,
            "close_cost": 0.0015,
            "min_cost": 5,
        },
    },
}

class WorkflowTest(TestCase):

    def setUp(self):
        super().setUp()

        provider_uri = "~/.qlib/qlib_data/cn_data"
        qlib.init(provider_uri=provider_uri, region=REG_CN)

        self.__my_prefix = 'my_'
        self.__rec_start_param = {'resume': True,
                                  'experiment_name': f"{self.__my_prefix}experiment",
                                  'recorder_name': f"{self.__my_prefix}recorder"}

        self._pkl_model = f"{self.__my_prefix}model.pkl"
        self._pkl_dataset = f"{self.__my_prefix}dataset.pkl"

    def test_backtest(self):
        with R.start(**self.__rec_start_param):
            recorder = R.get_recorder()

            pred = R.load_object('pred.pkl')
            placehorder_value = {"<PRED>": pred}
            pa_config = fill_placeholder(portfolio_analysis_config, placehorder_value)

            par = PortAnaRecord(recorder=recorder, config=pa_config)
            par.generate()


```

errors:

```
  File "/python3.12/site-packages/qlib/backtest/report.py", line 116, in _cal_benchmark
    raise ValueError(f"The benchmark {_codes} does not exist. Please provide the right benchmark")
ValueError: The benchmark ['SH000300'] does not exist. Please provide the right benchmark

```

DeBUG this method `qlib.utils.resam.get_higher_eq_freq_feature()`，Found that the method was called twice, the first time `freq=day` and the second time freq=1day`

pujitha24:
## Description

`get_higher_eq_freq_feature()` (in `qlib/utils/resam.py`) now normalizes the
`freq` argument (e.g. `"1day"` -> `"day"`) before querying `D.features`, so
that a freq string equivalent to but spelled differently than the freq used
to name feature files on disk still finds the data.

## Motivation and Context

Fixes https://github.com/microsoft/qlib/issues/1925

When a backtest executor is configured with `time_per_step="1day"`, that raw
string is threaded through to `PortfolioMetrics`/`Indicator` and on into
`_cal_benchmark()` in `qlib/backtest/report.py`, which calls
`get_higher_eq_freq_feature(..., freq="1day")`. Feature bin files on disk are
named using the *normalized* freq string (see
`FileFeatureStorage.__init__` in `qlib/data/storage/file_storage.py`, which
builds `file_name` from the raw freq, matching files like `$close.day.bin`).
Since `"1day"` and `"day"` are the same frequency but different strings, the
query for `"1day"` silently finds no data (it doesn't raise — the freq/support
check in `FileStorageMixin.uri` treats them as equal via `Freq.__eq__`, but
the literal filename lookup does not), and the resulting empty DataFrame
trips the `len(_temp_result) == 0` check, surfacing as the misleading error
`ValueError: The benchmark [...] does not exist. Please provide the right
benchmark` — even though the benchmark data is present under `day` freq.

Note this is a silently-empty-query / misleading-error bug, not a crash: the
existing `try/except (ValueError, KeyError)` fallback in
`get_higher_eq_freq_feature` was never reached here because no exception was
raised in the first place.

## How Has This Been Tested?

- [x] Added a targeted regression test, `tests/misc/test_resam.py`, which
  mocks `qlib.data.data.D` and asserts `get_higher_eq_freq_feature(freq="1day")`
  queries with the normalized `freq="day"`. Verified with `git stash` that
  this test fails on the pre-fix code (`AssertionError: '1day' != 'day'`) and
  passes after the fix.
- [x] `cd tests && python -m pytest misc/ -v` — 13 passed (includes the 2 new
  tests plus all pre-existing tests in that directory; no regressions).
- [x] `black qlib/utils/resam.py tests/misc/test_resam.py -l 120 --check --diff` — clean.
- [x] `pylint` (repo's configured ignore list) on `qlib/utils/resam.py` — 10.00/10.
- [x] `flake8` (repo's configured ignore list) on `qlib/utils/resam.py` — clean.
- [x] `mypy qlib/utils/resam.py` — 2 pre-existing errors reported, both on lines
  untouched by this change (unrelated to the fix).
- Not run: the full CI matrix (`make dev`, full `pytest tests/` against
  downloaded CN market data, sphinx docs, nbconvert) — this requires
  downloading several GB of market data and was impractical to run in this
  environment. The change is a small, self-contained string-normalization fix
  fully exercised by the targeted unit test above without needing live market
  data.

1. Pipeline test: not run (see "Not run" note above).
2. Your own tests: `python -m pytest tests/misc/test_resam.py -v` — 2 passed
   (fails on pre-fix code, passes after the fix).

## Types of changes
- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation

Fixes #1925


pujitha24:
Just checking in on this — it's still green and rebased on main; happy to make any changes if something would help move review along.

Issue #2352, opened 2026-09-20
IdxMax and IdxMin return the position of NaN instead of the extreme value


LudwigJMarx:
## 🐛 Bug Description

`IdxMax` and `IdxMin` return the position of the extreme value inside the
rolling window. Both use `ndarray.argmax` / `ndarray.argmin`, which do not skip
NaN: as soon as the window holds a NaN, they return its position instead.

`Max` and `Min` go through `pandas.rolling.max/min` and do skip NaN, so the two
contradict each other on the same series. `Mad` and `Rank` in the same file
filter NaN explicitly (`x[~np.isnan(x)]`); `IdxMax` and `IdxMin` do not.

This reaches Alpha158. `qlib/contrib/data/loader.py` builds three field
families on these two operators:

| field | expression | meaning per the comment in the loader |
|---|---|---|
| `IMAX{d}` | `IdxMax($high, d)/d` | days since the last high (Aroon) |
| `IMIN{d}` | `IdxMin($low, d)/d` | days since the last low (Aroon) |
| `IMXD{d}` | `(IdxMax($high,d)-IdxMin($low,d))/d` | distance between the two |

`$high` and `$low` are NaN while an instrument is suspended.

## To Reproduce

1. Run the script below on a clean checkout. It only replaces the data access
   of a leaf feature; the operators under test run their real code.

```python
import numpy as np
import pandas as pd
from qlib.data.base import Feature
from qlib.data.ops import IdxMax, Max


class SeriesFeature(Feature):
    def __init__(self, name, values):
        super().__init__(name)
        self._values = values

    def _load_internal(self, instrument, start_index, end_index, freq):
        return pd.Series(self._values, dtype="float64")


values = [1.0, 5.0, np.nan, 2.0, 3.0]
window = 3
idx = IdxMax(SeriesFeature("a", values), window).load("x", 0, len(values) - 1, "day")
mx = Max(SeriesFeature("b", values), window).load("x", 0, len(values) - 1, "day")

for i in range(len(values)):
    w = values[max(0, i - window + 1) : i + 1]
    print(f"{w}  Max={mx.iloc[i]}  IdxMax={idx.iloc[i]}  points at {w[int(idx.iloc[i]) - 1]}")
```

2. Output on `be725493`:

```
[1.0]  Max=1.0  IdxMax=1.0  points at 1.0
[1.0, 5.0]  Max=5.0  IdxMax=2.0  points at 5.0
[...]

PR #2353, opened 2026-09-20
fix(ops): make IdxMax and IdxMin skip NaN inside the window

LudwigJMarx:
## Description

`IdxMax` and `IdxMin` used `ndarray.argmax` / `ndarray.argmin`, which return
the position of a NaN rather than skipping it. Replaced with `np.nanargmax` /
`np.nanargmin`, four lines.

## Motivation and Context

Refs #2352.

`Max` and `Min` skip NaN through pandas, `IdxMax` and `IdxMin` did not, so the
two disagreed on the same series. `Mad` and `Rank` in the same file already
filter NaN explicitly.

Alpha158 builds `IMAX{d}`, `IMIN{d}` and `IMXD{d}` on these operators over
`$high` and `$low`, which are NaN while an instrument is suspended.

No guard against `ValueError: All-NaN slice encountered` is needed, and that is
measured rather than assumed: with `min_periods=1` pandas does not call `apply`
at all for a window that holds only NaN. Checked for both the `rolling` and the
`expanding` path. The `min_periods=1` sits on the same line as the `apply`
call, so the assumption is local.

Rejected: filtering NaN out first, the way `Mad` and `Rank` do. The returned
position would then be the one inside the filtered array rather than inside the
window, which is exactly the quantity the field means (days since the high).

## How Has This Been Tested?

`tests/test_all_pipeline.py` is the one that matters here: it runs
`DATASET_ALPHA158_CLASS` over csi300 from 2008 to 2020, so it exercises the
fields this change touches, and `test_1_backtest` checks a real threshold.
It carries `pytest.mark.slow`, so CI skips it. Ran both sides:

| | result |
|---|---|
| before the fix, on `7bd57c18` | 3 passed in 357 s |
| after the fix, on `4f5941e0` | 3 passed in 258 s |

The runtimes differ because `GBDT_MODEL` sets no seed and uses
`num_threads: 20`. The test checks a threshold, not equality.

New file `tests/ops/test_rolling_operator.py`. It asserts no invented number
but the invariant against the real sibling operators: `IdxMax` has to point at
the value `Max` reports. The two commits carry the evidence: the test is red on
`7bd57c18` and green on `4f5941e0`.

Before the fix:

```
FAILED ops/test_rolling_operator.py::TestIdxRollingOperator::test_idxmax_points_at_the_maximum
FAILED ops/test_rolling_operator.py::TestIdxRollingOperator::test_idxmin_points_at_the_minimum
E   AssertionError: np.float64(nan) != np.float64(5.0) : IdxMax points at nan but Max reports 5.0
E   AssertionError: np.float64(nan) != np.float64(1.0) : IdxMin points at nan but Min reports 1.0
2 failed in 2.07s
```

After the fix:

```
2 passed in 2.21s
```

Rest of the suite, same command as CI (`cd tests && pytest . -m "not slow"`):

| | result |
|---|---|
| before, untouched tree | 71 passed, 1 skipped, 10 deselected, 79 subtests, 818 s |
| after | 73 passed, 1 skipped, 10 deselected, 79 subtests, 718 s |

The two extra are the new tests.

Counter-check that nothing else moves: 200000 random windows without NaN, ties
forced, `argmax` against `nanargmax` and `argmin` against `nanargmin`, zero
differences.

`black -l 120 --check` clean, `flake8` clean, `pylint` 10.00/10 on
`qlib/data/ops.py`.

Environment: Python 3.12.13, numpy 2.5.3, pandas 2.3.3, MacOS 26.6.2 arm64,
commit be725493.

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation


LudwigJMarx:

koriyoshi2041 (commented):
The sibling-operator invariant is a strong regression shape. One branch changed here is not pinned by the new tests, though: both `IdxMax` and `IdxMin` switch the `N == 0` expanding path to `nanarg*`, while `TestIdxRollingOperator.WINDOW = 3` only exercises rolling.

I replayed the same series with `expanding(min_periods=1)`: the old `argmax`/`argmin` path points at NaN from indices 2–4, while the proposed `nanargmax`/`nanargmin` path keeps pointing at the sibling `Max`/`Min` value. I also confirmed pandas does not invoke the callback for an all-NaN expanding prefix, matching the PR's stated safety assumption.

Could the regression cover both `N=3` and `N=0` (for example by running the same invariant helper for each)? That would pin the second changed execution path and the expanding-specific all-NaN assumption rather than leaving them supported only by the PR description.

LudwigJMarx:
Confirmed, and the test did not catch it. I reverted the expanding branch to `argmax`/`argmin`, left the rolling branch fixed, and ran the tests: both stayed green. The mutation survived, so the test pinned only half of the diff.

`WINDOWS = (3, 0)` now runs the same invariant helper for both branches, with a `subTest` per window and index.

Mutation results:

| mutation | before | after |
|---|---|---|
| expanding branch back to `argmax`/`argmin` | 2 passed | 6 failed |
| rolling branch back to `argmax`/`argmin` | 2 failed | 6 failed |
| no mutation | 2 passed | 2 passed, 20 subtests |

The expanding path, measured against the real operators on `[1.0, 5.0, nan, 2.0, 3.0]`:

| i | prefix | old `IdxMax` points at | new points at | `Max` |
|---|---|---|---|---|
| 0 | [1.0] | 1.0 | 1.0 | 1.0 |
| 1 | [1.0, 5.0] | 5.0 | 5.0 | 5.0 |
| 2 | [1.0, 5.0, nan] | nan | 5.0 | 5.0 |
| 3 | [1.0, 5.0, nan, 2.0] | nan | 5.0 | 5.0 |
| 4 | [1.0, 5.0, nan, 2.0, 3.0] | nan | 5.0 | 5.0 |

`IdxMin` behaves the same way. The all-NaN assumption also holds on the expanding path: with `min_periods=1`, pandas does not call `apply` for a prefix that holds only NaN.

Rest of the suite on `3d5c9906`: 73 passed, 1 skipped, 10 deselected, 99 subtests. The reference run on the untouched tree was 71 passed with 79 subtests, so the 20 extra subtests are the new cases.

koriyoshi2041 (approved):
The updated regression now covers both changed branches (`N=3` rolling and `N=0` expanding) and catches the old `arg*` implementation in either branch. I reran the exact head locally: 2 tests / 20 subtests pass, and `git diff --check` is clean.

Issue #2354, opened 2026-09-20
collect_info.py does not start: pkg_resources is gone from setuptools 82


LudwigJMarx:
## 🐛 Bug Description

`scripts/collect_info.py` aborts on its first import. This is the script the
bug report template points users to for their environment details.

Two causes, one behind the other, both in the same file:

| | cause | effect |
|---|---|---|
| 1 | `import pkg_resources` at module level | `ModuleNotFoundError`, the script never starts |
| 2 | `REQUIRED` lists `wheel` and `cython` | `DistributionNotFound`, the script dies part way through its output |

The second one only becomes visible once the first is fixed. `wheel` and
`cython` sit in `build-system.requires`, not in `dependencies`, so a plain
`pip install pyqlib` does not have them.

## To Reproduce

1. Install qlib in an environment with setuptools 82 or newer. Since Python
   3.12 `python -m venv` no longer seeds setuptools at all, so `pkg_resources`
   can be missing regardless of any version boundary.
2. `cd scripts && python collect_info.py all`
3. Output:

```
Traceback (most recent call last):
  File "scripts/collect_info.py", line 5, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

Measured boundary, one throwaway venv per version, nothing else changed:

| setuptools | ships `pkg_resources` |
|---|---|
| 80.9.0 | yes |
| 81.0.0 | yes, with a deprecation warning |
| 82.0.0 | no |

## Expected Behavior

`collect_info.py all` prints the system, Python and package details and exits
with 0, including when a package from its list is not installed.

## Environment

Collected by hand, since the script itself does not run:

 - Qlib version: 0.9.8.dev34 (editable install)
 - Python version: 3.12.13
 - OS: MacOS 26.6.2, arm64
 - Commit number: be725493eb1a6bbb42bf11b37aa7669f59610ff1
 - setuptools 84.0.0

## Additional Notes

#1947 covers the same line, but from June 2025, when it was still a
`UserWarning` with a different traceback behind it. Today the script does not
start at all.

PR #2355, opened 2026-09-20
fix(scripts): replace pkg_resources in collect_info with importlib.metadata

LudwigJMarx:
## Description

`scripts/collect_info.py` used `pkg_resources`, which setuptools stopped
shipping with 82.0.0, and it treated every name in `REQUIRED` as installed.
Switched to `importlib.metadata` and made a missing package print a line
instead of raising.

## Motivation and Context

Refs #2354.

The bug report template points users at this script for their environment
details, so it fails exactly when someone is trying to file a report.

`importlib.metadata` has been in the standard library since Python 3.8 and
`requires-python` is `>=3.8.0`, so this adds no dependency.

Two causes, and the second is only reachable once the first is fixed:
`import pkg_resources` stops the script before it starts, and `REQUIRED` lists
`wheel` and `cython`, which sit in `build-system.requires` rather than in
`dependencies`. Both stand in the way of the same outcome, so they are handled
together rather than in two PRs.

Rejected: dropping `wheel` and `cython` from `REQUIRED`. For a diagnostic
script the useful answer is that a package is missing, not an abort, and the
same handling covers later edits to that list.

## How Has This Been Tested?

New file `tests/test_collect_info.py`. The result must not depend on whichever
setuptools the machine happens to hold, so the test runs the script in a
subprocess behind a `meta_path` finder that blocks `pkg_resources`. That keeps
it red on any setuptools version while the import is in the file. The two
commits carry the evidence: red on `8c7e8698`, green on `38a30864`.

Before the fix:

```
FAILED test_collect_info.py::TestCollectInfo::test_runs_without_pkg_resources
FAILED test_collect_info.py::TestCollectInfo::test_reports_a_package_that_is_not_installed
E   AssertionError: 1 != 0 : collect_info.py all failed:
E     File "collect_info.py", line 5, in <module>
E       import pkg_resources
E   ImportError: pkg_resources is not available
2 failed in 2.05s
```

After the fix:

```
2 passed in 1.61s
```

And by hand, the way the template asks for it
(`cd scripts && python collect_info.py all`), exit 0:

```
Qlib version: 0.9.8.dev34
setuptools==84.0.0
wheel==0.48.0
cython: not installed
pyyaml==6.0.3
...
setuptools-scm==10.2.3
```

Counter-check that the replacement reports the same values: in a venv on
setuptools 81, where both APIs still exist,
`pkg_resources.get_distribution(x).version` against
`importlib.metadata.version(x)` over seven packages including `ruamel.yaml`,
`python-redis-lock`, `pydantic-settings` and `setuptools-scm`: zero
differences.

Rest of the suite, same command as CI (`cd tests && pytest . -m "not slow"`):

| | result |
|---|---|
| before, untouched tree | 71 passed, 1 skipped, 10 deselected, 79 subtests |
| after | 73 passed, 1 skipped, 10 deselected, 79 subtests |

The two extra are the new tests.

`black -l 120 --check` clean, `pylint` 10.00/10 on `scripts/collect_info.py`.

Environment: Python 3.12.13, setuptools 84.0.0, MacOS 26.6.2 arm64, commit
be725493.

## Types of changes

- [x] Fix bugs
- [ ] Add new feature
- [ ] Update documentation


koriyoshi2041 (approved):
The stdlib replacement preserves the existing output shape for installed distributions and, importantly, turns absent build-only packages into useful diagnostics instead of aborting the report.

I checked exact head `38a30864`: the focused tests pass (2/2), `git diff --check` is clean, and `python scripts/collect_info.py all` exits successfully in a Python 3.12 environment where both `wheel` and `cython` are absent and reports each as `not installed`.

LudwigJMarx:
While looking at the same question on #2353, I found the same shape here, and the test did not catch it either.

`test_reports_a_package_that_is_not_installed` relied on `cython` being absent, with a `skipIf` for machines that have it. On such a machine the test is skipped, so the `except PackageNotFoundError` branch runs without a test.

Measured, not assumed. A `cython` dist-info on the path makes all 26 names in `REQUIRED` resolve. In that state, with a bare `raise` in place of the except branch:

| mutation | without cython | with cython |
|---|---|---|
| except branch to `raise`, before | 2 failed | **1 passed, 1 skipped** |
| except branch to `raise`, now | 2 failed | 2 failed |
| back to `pkg_resources`, now | 2 failed | 2 failed |

The driver now makes one name missing, the same way it already blocks `pkg_resources`: it wraps `importlib.metadata.version` and raises `PackageNotFoundError` for `cython`. The `skipIf` is gone and both branches run everywhere.

The point raised in the review on #2353 applies to this PR as well.

Rest of the suite on `9d7187a5`: 73 passed, 1 skipped, 10 deselected, 79 subtests, the same as before.

